Issue #3, opened 2018-12-07
Tip frequencies are lacking


trvrb:
We need to calculate `tip-frequencies.json` via KDE frequencies in order to restore the frequencies panel. I believe this functionality should be part of `augur frequencies` where `--tree` is passed in. And then maybe have flags for `--method kde` and `--tips-only` or some such.

rneher:
sounds good to me. I haven't touched the KDE part to avoid messing with John's dev. i would make integration easier if the KDE part was sub-classed in a similar way into a part that operates on alignments and trees with an underlying core machinery. @huddlej, probably best if you hash out a proposal for this.

huddlej:
Resolved by #7.

PR #7, opened 2018-12-15
Build tip frequencies for auspice

huddlej:
Addressing issue #3, this PR modifies auspice configuration to hide the map and show the frequencies panel and uses the KDE frequencies method to estimate region-weighted frequencies from
the refined tree.


huddlej:
Merging this since builds worked for @barneypotter24 and python 2/3 compatibility has been worked out.

PR #9, opened 2018-12-18
Annotate tree nodes with titer substitution model

### LINKED ISSUE #4: cTiterSub is lacking
Currently `augur titers sub` outputs a JSON that has entries for `avidity`, `potency`, `substitution`, `titers`, but lacks a `nodes` entry that could be used to decorate the final `tree.json` via `augur export`.

This change needs to happen to `augur titers sub` and then the seasonal build can be updated.

huddlej:
Passes a tree to the titer substitition model to add a "nodes" attribute to the
JSON output with antigenic advance values per node. Additionally, adds the
resulting JSON to the default inputs for auspice export and adds a corresponding
auspice color-by configuration for "cTiterSub".

Relies on code from the corresponding `ctitersub-tree` branch of augur.

Resolves #4.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -347,7 +347,8 @@ rule titers_sub:
     input:
         titers = rules.download_titers.output.titers,
         aa_muts = rules.translate.output,
-        alignments = translations
+        alignments = translations,
+        tree = rules.refine.output.tree
     params:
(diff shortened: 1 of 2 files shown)


rneher:
this looks good.

PR #10, opened 2018-12-20
Align sample filtering requirements with previous augur builds

### LINKED ISSUE #6: Consistently filter sequences using augur filter
To make these flu builds consistent with previous augur builds, I'm suggesting the following modifications to the current Snakefile:

 - Use `augur filter` to filter sequences by length and outlier status prior to selecting strains
 - Modify `scripts/select_strains.py` to take both sequences and metadata as inputs and to output selected sequences
 - Remove current Python-only `filter` rule from the Snakefile (it is no longer needed after the aforementioned modifications)

If `select_strains.py` outputs filtered sequences themselves instead of strain names, that will remove the need for an `--include-only` type of option for the augur filter command and make the flu-specific script more consistent with existing augur command rather than the other way around.

huddlej:
Adds a new filter rule that applies the augur filter command to omit sequences
shorter than a minimum length, outliers, and samples with missing geographic
metadata prior to strain selection, in an attempt to match the filtering
requirements used by previous augur builds [1]. This commit also updates
`select_strains.py` to accept one or more FASTA files and to only consider
metadata for samples whose sequences made it through the aforementioned
filters. After strains are selected, we still need a rule to extract the segment
sequences corresponding to each strain. This rule has been renamed from "filter"
to "extract" and its body has been replaced with a call to a standalone script
that performs the extraction using the same logic as the original Python code.

On a related note, this commit also imposes an upper date limit on included
strains in `select_strains.py` to prevent reference strains from the future from
being included in a build. This modification has no effect on standard "live"
builds, but it ensures consistent output for custom builds that request a
specific time interval from the past (e.g., 2005-2015) that should not be
assigned any sequences after the last date in the requested time interval.

Resolves #6.

[1] https://github.com/nextstrain/augur/blob/3.1.0/builds/flu/flu.prepare.py#L102-L112

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -168,7 +168,6 @@ rule filter:
         Filtering {wildcards.lineage} {wildcards.segment} sequences:
           - less than {params.min_length} bases
           - outliers
-          - samples with missing region and country metadata
         """
     input:
         metadata = rules.parse.output.metadata,
@@ -185,7 +184,6 @@ rule filter:
             --metadata {input.metadata} \
             --min-length {params.min_length} \
             --exclude {input.exclude} \
-            --exclude-where country=? region=? \
             --output {output}
         """
 
--- a/Snakefile
+++ b/Snakefile
@@ -3,6 +3,7 @@ import pandas as pd
 from treetime.utils import numeric_date
 
 path_to_fauna = '../fauna'
+min_length = 900
 segments = ['ha', 'na']
 lineages = ['h3n2']
 resolutions = ['2y', '3y', '6y', '12y']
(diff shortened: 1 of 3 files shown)


Snakefile
```diff
@@ -3,6 +3,7 @@ import pandas as pd
 from treetime.utils import numeric_date
 
 path_to_fauna = '../fauna'
+min_length = 900
```
rneher: def min_length(w):
   segment_min_length = {'ha':900, 'na':1300, 'PB1':2000', 'PB2':2000}
   return segment_min_length.get(w.segment, 900)

scripts/select_strains.py
```diff
@@ -177,6 +181,19 @@ def parse_metadata(segments, metadata_files):
         metadata[segment] = tmp_meta
     return metadata
 
+def parse_sequences(segments, sequence_files):
+    """Load sequence names into a dictionary of sets indexed by segment.
+    """
+    sequences = {}
+    for segment, filename in zip(segments, sequence_files):
+        sequence_set = Bio.SeqIO.parse(filename, "fasta")
+        sequences[segment] = set()
```
rneher: wouldn't:
sequences[segment] = set(Bio.SeqIO.parse(filename, "fasta"))

do the trick? Does it matter with list of set?

rneher: sorry, I misread and didn't realize that you only glob the names. still a list comprehension would do.

scripts/select_strains.py
```diff
@@ -206,28 +224,44 @@ def parse_metadata(segments, metadata_files):
...
+    sequence_names_by_segment = parse_sequences(args.segments, args.sequences)
+
     # read in meta data, parse numeric dates
     metadata = parse_metadata(args.segments, args.metadata)
+
+    # eliminate all metadata entries that do not have sequences
+    filtered_metadata = {}
+    for segment in metadata:
+        filtered_metadata[segment] = {}
+        for name in metadata[segment]:
```
rneher: since you have the names as sets already, you could use set intersections. but just a matter of style (and speed, but irrelevant here).

trvrb:
This overall behavior looks good to me John. I just removed the exclusion of samples without geo information. I'd like to have these flagged as part of the build so that we can go in an repair them in fauna. I pushed a change to auspice so it shouldn't break due to lack of lat/long information for all demes: https://github.com/nextstrain/auspice/commit/6bc9025d0d15b725f78e1ae9cbb48e0031eba22d.

trvrb:
I'm going to merge this now, despite @rneher's requests. I'm working on filtering now and I'd like to build on top of this. I'm hoping that @rneher's requests can be made to `master` or via another PR.

PR #14, opened 2019-01-03
Import clade membership using HA tree structure

### LINKED ISSUE #13: Identify internal nodes with the HA tree when importing tip clades
The current script to import tip clades from HA to non-HA segments identifies internal nodes by [checking that the first four characters of the node name match "NODE"](https://github.com/nextstrain/seasonal-flu/blob/1f07773/scripts/import_tip_clades.py#L31). The fact that internal nodes have this specific name is [a design decision from TreeTime](https://github.com/neherlab/treetime/blob/273a236/treetime/treeanc.py#L766). This approach tries to infer the data structure from the data contents when information about the structure is already available in an external Newick file.

I'd suggest we pass the original HA tree Newick to this script and use the Bio.Phylo `is_terminal` method to identify internal nodes instead of testing for specific text. This change will prevent the script from silently breaking when/if TreeTime names internal nodes something different in the future and will avoid confusion when someone would otherwise have to figure out where the "NODE" text comes (I admit this took me a little while to track down). Although this change couples the script to the Bio.Phylo interface, that interface is something more easily grepped for and replaced during a future refactor.

The Newick tree for the current segment is [already an input for the clades rule](https://github.com/nextstrain/seasonal-flu/blob/1f07773/Snakefile#L515-L538), so it would not be difficult to force that input tree to always be the HA tree instead.

huddlej:
Uses the original HA tree used to assign clade memberships to tips to select
only tips when importing clade memberships for other segments. Note that this
commit omits export of any internal node data, since we would not expect
internal node names to match meaningfully between trees for different segments
and auspice can handle missing data for internal nodes gracefully.

Closes #13.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -515,7 +515,7 @@ rule tree_frequencies:
 rule clades:
     message: "Annotating clades"
     input:
-        tree = rules.refine.output.tree,
+        tree = "results/tree_{lineage}_ha_{resolution}.nwk",
         nt_muts = rules.ancestral.output,
         aa_muts = rules.translate.output,
         clades = _get_clades_file_for_wildcards
@@ -533,6 +533,7 @@ rule clades:
         else:
             shell("""
                 python scripts/import_tip_clades.py \
+                    --tree {input.tree} \
                     --clades {input.clades} \
                     --output {output.clades}
             """)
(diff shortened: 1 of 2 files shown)


huddlej:
This PR conflicts with the generalized builds branch, so I can update this PR when that branch is merged or vice versa.

rneher:
We could just merge this now and I merge master into 'generalized...' first thing tomorrow. The modifications in `select_strains.py` should be fine, its just the tree filename, right?

huddlej:
Ok, yes, just the tree filename should cause a conflict. Merging now.

PR #15, opened 2019-01-08
Generalized builds

rneher:


trvrb:
Thank you for all the work here @rneher. This worked really for me. There is some more cleanup I'd like to do, but quite happy with the general approach.

PR #16, opened 2019-01-08
Add selection flag for segment.

barneypotter24:
This PR adds a selection flag for segment to the batch submission script for seasonal flu. This gives the user control over all three variables (segment, lineage, resolution) as opposed to just two. Allows re-runs (i.e. for an added outlier) to ignore NA builds, as is often the case.

Defaults to ["ha", "na"].

### DIFF
--- a/batch.py
+++ b/batch.py
@@ -11,23 +11,14 @@ import os
 
 if __name__ == '__main__':
 
-    parser = argparse.ArgumentParser(description='Run flu builds on aws')
-    parser.add_argument('-l', '--lineages', nargs='+', type = str,  help ="flu lineages to include")
-    parser.add_argument('-r', '--resolutions', nargs='+', type = str,  help ="flu resolutions to include")
-    parser.add_argument('-s', '--segments', nargs='+', type = str, help ="flu segments to include")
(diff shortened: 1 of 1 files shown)


trvrb:
This looks good. Thanks Barney.

batch.py
```diff
@@ -14,13 +14,15 @@
     parser = argparse.ArgumentParser(description='Run flu builds on aws')
     parser.add_argument('-l', '--lineages', nargs='+', type = str,  help ="flu lineages to include")
     parser.add_argument('-r', '--resolutions', nargs='+', type = str,  help ="flu resolutions to include")
+    parser.add_argument('-s', '--segments', nargs='+', type = str, help ="flu segments to include")
```
huddlej: If you include `defaults = ["ha", "na"]` here, you can remove the check for `params.segments is None` below. This will also expose the default segments to the user on the command line, in the future when the ArgumentParser's formatter class is set to `argparse.ArgumentDefaultsHelpFormatter`.

barneypotter24: Thanks for the suggestion @huddlej. I did it this way to follow the standard that already existed for `lineages` and `resolutions`. Changing now for all three.

trvrb: Good call.

PR #17, opened 2019-01-08
Add outliers for rebuilds.

barneypotter24:
Merge in additional outliers from dcd019c to run rebuilds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,5 +42,8 @@ ehthumbs.db
 Thumbs.db
 *~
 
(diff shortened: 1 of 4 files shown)


barneypotter24:
This was intentional (but maybe dumb). My goal was to commit updates to outliers on `master`, but I wanted those same outliers to be usable from the unmerged `segment-selection` so that I could run builds from that branch with the new outliers added. Is this an incorrect thing to do?

PR #18, opened 2019-01-10
Ensure no future sequences are included in builds

huddlej:
Fixes a bug in the logic associated with filtering strains by date which was causing strains from outside requested time intervals to be included in the final output. This bug was due to an off-by-one error in parsing of months where metadata month values were zero-indexed and being compared to one-indexed months in datetime objects.

This commit addresses this error by filtering strains by numeric float dates at the beginning of the flu subsampling function. This approach allows users to specify time intervals to the day and have the subsampling respect that precise of a request whereas the previous implementation with the `cat_valid` function only checked the year and month of strains.

Note: the bug described here is only an issue when the `--time-interval` option is used to specify a date other than the current date. Standard use with the `--resolution` option is not affected by this bug.

### DIFF
--- a/scripts/select_strains.py
+++ b/scripts/select_strains.py
@@ -81,6 +81,16 @@ def populate_categories(metadata):
 
 
 def flu_subsampling(metadata, viruses_per_month, time_interval, titer_fname=None):
+    # Filter metadata by date using the given time interval. Using numeric dates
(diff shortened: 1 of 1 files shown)


trvrb:
This is great John. Thanks for the catch.

PR #23, opened 2019-02-07
Export root sequences JSON as a standard auspice output

huddlej:
This export costs little in time and space, but it enables downstream analyses that depend on amino acid sequences for seasonal flu. For example, auspice JSONs with the root sequence could be used as inputs to forecasts without needing to build new trees just for that purpose.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -58,7 +58,7 @@ rule export:
     output:
         auspice_tree = "auspice/flu_{center}_{lineage}_{segment}_{resolution}_{passage}_{assay}_tree.json",
         auspice_meta = "auspice/flu_{center}_{lineage}_{segment}_{resolution}_{passage}_{assay}_meta.json",
-        auspice_seq = "auspice/flu_{center}_{lineage}_{segment}_{resolution}_{passage}_{assay}_seq.json"
+        auspice_seq = "auspice/flu_{center}_{lineage}_{segment}_{resolution}_{passage}_{assay}_root-sequence.json"
     shell:
         """
         augur export \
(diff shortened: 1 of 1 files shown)


trvrb:
@huddlej ---

I believe that `flu_seasonal_h3n2_ha_3y_seq.json` should not live in `auspice/`. This `auspice/` directory is directly used as input for auspice via locally running `auspice` or by uploading to s3 via `nextstrain deploy ... auspice/flu_seasonal_*`. Auspice currently does not use these files. I would definitely think that `results/flu_seasonal_h3n2_ha_3y_seq.json` would be a better place for this if you need it for downstream fitness forecasts.

Although I really don't understand why this is part of `augur export`. Am I confused and auspice can now handle these `seq.json` files?

jameshadfield:
>Am I confused and auspice can now handle these `seq.json` files

Not yet -- but my memory of this conversation a few months back is that one day auspice will be able to use these files to display non-polymorphic bases in the entropy plot. At that time we discussed incorporating this in the main JSON file and decided to use a separate file instead.

huddlej:
Here is [the original PR adding the root sequence to auspice export](https://github.com/nextstrain/augur/pull/212) and [some of the related discussion on Slack](https://bedfordlab.slack.com/archives/C131TERB5/p1537213458000100?thread_ts=1537213458.000100).

My hope was to make running forecasts as painless as possible for future users. If a user already had a model in hand (a list of predictors and their parameter values or a JSON file with that information), they technically only need the auspice tree, meta, and sequence JSONs to make forecasts. Having these three JSON files on data.nextstrain.org would make creating forecasts as simple as downloading the latest build JSONs and running a script on the command line from any computer without needing access to the original build directory (on rhino, AWS, the laptop of the person who ran the build, etc.).

Having these JSONs available to auspice could also enable embedding forecasts in auspice panels (calculating the projections from the model params and JSONs on the fly in Javascript).

trvrb:
Thank you John!

PR #29, opened 2019-04-29
Update reference to titer TSV file in Travis config

huddlej:
The titer TSV was renamed but the Travis script was not updated. This led CI
builds to fail when titers were missing and could not be downloaded from
RethinkDB.


huddlej:
Travis builds are currently failing since augur has been updated to the new `augur distance` command. This commit fixes an issue that predated the new augur version that was causing Travis builds to fail when looking for titer TSVs.

PR #30, opened 2019-04-29
Add logs and data directories to clean.

barneypotter24:
This is a QOL change since before each new build I had to run additional `rm` commands for the `log` and `data` directories after running `snakemake clean`. This adds those directories to the regular command.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -109,6 +109,8 @@ rule clean:
         "results ",
         "targets ",
         "auspice ",
-        "auspice-who"
+        "auspice-who ",
+        "logs ",
+        "data"
     shell:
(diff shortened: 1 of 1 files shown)


trvrb:
@barneypotter24: Generally, I prefer to keep `data/` out of 'clean'. This mirrors general 'clean' vs 'clobber' practices. Things in `data/` take a while to redownload. And for other people they may not have replacements. We can make a 'clobber' rule as well that includes `data/`.

rneher:
I'd prefer `clean-all` over `clobber`, but I agree with keeping data out of simple `clean`.

PR #34, opened 2019-07-02
Codon align

trvrb:


trvrb:
This is super clever (as always) @rneher. The approach makes sense to me and this seems to work well across existing builds.

PR #36, opened 2019-08-21
Add support for forecasting live builds

huddlej:
Includes construction of a tip attributes table, calculation of pairwise
distances between amino acid sequences, forecasting of tips using a fixed model,
and inclusion of forecasted frequencies in the auspice tip frequencies JSON with
a new "projection_pivot" key to enable auspice to customize its display of
future frequencies.

Forecast support relies on a flu-forecasting submodule that will eventually be
replaced by an external Python module dependency.


trvrb:
I've elected to copy files over from `flu-forecasting`. I believe this should prove less annoying than a git submodule. Files from `flu-forecasting` that were necessary:
```
scripts/
  calculate_delta_frequency.py
  calculate_target_distances.py
  convert_translations_to_json.py
  node_data_to_table.py
src/
  fit_model.py
  forecast_model.py
  weighted_distances.py
  forecast/
    fitness_model.py
    fitness_predictors.py
    metrics.py
```

trvrb:
This branch failed Travis CI because of an upstream issue in old `master` resolved by https://github.com/nextstrain/seasonal-flu/commit/e30af65088036fdaa996bd08d9353a2c7274c7e6. The PR (after merging) is passing.

PR #40, opened 2019-11-13
Vaccine info node-data JSONs

jameshadfield:
The information, previously exported in `export [v1]` via the auspice config file are stored in dedicated JSON files for `export v2`. The auspice config files remain unchanged so that `export v1` still works as expected, but this information should be removed from them when it's no longer needed.

You will need to use this PR of augur: https://github.com/nextstrain/augur/pull/410

P.S. I don't believe the travis failures are due to this PR, although I only tested it on `3y` builds to speed things up.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -46,7 +46,8 @@ def _get_node_data_for_export(wildcards):
         rules.titers_sub.output.titers_model,
         rules.clades.output.clades,
         rules.traits.output.node_data,
-        rules.lbi.output.lbi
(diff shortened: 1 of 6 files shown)


trvrb:
This looks excellent James. Thank you! I didn't realize that you can just nest attributes in node data JSONs. Very clean. I tested and this works for me.

The Travis failures are due to running on latest Augur release v5.4.1. The `export-v2` branches (in this and other pathogen repos) should start passing when Augur v6 is released.

PR #45, opened 2019-12-18
Update WHO build

joverlee521:
This PR fixes the downstream errors in the WHO build due to changes in augur.

1. Ignore the top level `generated_by` key within the frequencies JSON generated by `augur frequencies`.

2. Remove lowering and conversion to snakecase of country names from the static vaccine TSVs, since the metadata now comes in with the country names already "prettified".


trvrb:
Looks great. Thanks Jover.

PR #47, opened 2020-01-09
Update forecasts to use best model from validation and testing

huddlej:
This PR updates forecasts to use the best model from validation and testing which was HI cross-immunity and mutational load. This update required the addition of several new scripts and rules to calculate cross-immunity. Switching models also revealed a couple issues with the original forecast integration that are resolved here. These include:

  - no longer censoring strains sampled in the last month from frequency estimation (this censoring caused recent strains to be omitted from projections of the future population)
  - decoupling of forecasts and weighted distance to the future calculations (now all tips get a weighted distance to the future even if they are not included in the forecast)
  - forecasts were not described anywhere and now are briefly explained below the acknowledgments
  - the bandwidth used for KDE frequency estimation is now set to the optimal 2-month value

Integrating this new model also revealed issues with some of the original model code that never appeared in training/validation data. This PR fixes those issues and also dramatically speeds up pairwise tree distance calculations.

Other recent changes include:

  - add a color-by attribute for number of HI titers per tip (shown below) to clarify data available to the antigenic advance and novelty views
  - define a formal clade for A1b/94N
  - support forecasting with multiple models
  - use "WHO" builds instead of "CDC" builds for forecast tables such that all available titers are used in forecasts

image

See [an example 2y build on the staging server](https://nextstrain.org/staging/flu/who/h3n2/ha/2y/cell/hi?d=tree,frequencies&p=full), to try out the new color by options and view projections from the new model.


huddlej:
A couple recent changes are worth review/discussion:

  - commit 400360d changes the Snakemake workflow for building auspice tip frequencies such that forecasts can be built for multiple models and the tip frequencies JSONs are copied instead of moved. This latter change is important to prevent Snakemake from constantly rebuilding files that already exist and for picking the forecasts for the best model.
 - commit 79e31a5 defines a clade for A1b/94N which may not be something we want, but since it is such a large component of A1b/131K's potential success I wanted to represent it here and in the forecasts

huddlej:
Thank you, @rneher! I've removed titer counts per tip as a display option and also confirmed that 2y builds work for H3N2, H1N1pdm, Vic, and Yam with this branch and also that a 6y H3N2 build works as expected.

I've also confirmed that a handful of the WHO builds run and load as expected in modern auspice. I haven't tried them with the old auspice though (due to Jekyll dependency hell).

huddlej:
Merging this now to prevent diverging from master, too much. The current state of this PR fulfills original requests for changes by @rneher and has been tested locally. @joverlee521 will test further with live builds.

PR #59, opened 2021-03-22
Annotate recency of submission dates

### LINKED ISSUE #55: Color strains by database submission date
Strain submission to GISAID and other databases often lags by one to two months from the actual collection date. When we make new seasonal flu builds, it would be helpful to identify which strains were recently added to the database.

The [ncov builds address this problem by adding a "recency" color-by attribute](https://github.com/nextstrain/ncov/blob/22c435106b39f6b644ee9b80208752a9612436e8/workflow/snakemake_rules/main_workflow.smk#L750-L764). This [categorical attribute bins submission dates into useful groups for ncov analysis](https://nextstrain.org/ncov/global?c=recency).

For flu, a similar categorical attribute that highlights sequences submitted in the last month or before should be sufficient.

huddlej:

Adds a script, configs, and Snakemake rules to annotate how recently each strain's sequence was submitted to its respective database based on today's date. The new script borrows from [the ncov workflow's script to construct recency annotations](https://github.com/nextstrain/ncov/blob/master/scripts/construct-recency-from-submission-date.py) but implements a more flexible user interface for defining and labeling the intervals to bin sequences into.

The resulting "recency" annotation appears in Nextstrain builds as "Submission date" in the color-by and filter fields.

### Related issue(s)  

Fixes #55

### Testing

I tested this locally with a small 2y H3N2 HA tree (using 5 viruses per month) and produced the following output:

Example recency annotation for a small 2y H3N2 HA tree

I also tested this PR with full 2y builds of all four lineages and had no issues.


rneher (approved):
Thanks, John! this looks good to me!

trvrb:
This looks great to me as well John. I think this can be merged from my perspective.

PR #73, opened 2021-12-03
Update clades to match WHO CC recommendations

huddlej:
 
Updates clade definitions for all four major lineages based on recent WHO CC recommendations. Clade definitions may deviate from official recommendations when amino acid substitutions in the recommendations do not form a monophyletic group or do not match any mutations present in a 12y flu tree for the lineage. In the first case, the clade definitions maintain the original Nextstrain nucleotide markers for clades. In the second case, the definitions use the most likely corrected substitutions.

Clades for H1N1pdm that precede the earliest WHO CC recommendations (pre-6B.1 and 6B.2) retain the original Nextstrain numbers, to distinguish pandemic-era clades from more recently circulating clades.

Below are lineage-specific notes about the clade name changes.

### H3N2

* 3c3.A -> added 3c3.A1 based on A/Kansas/14/2017
* A1b/135N -> dropped as a “no call (old virus)” by CCs
* A2 and A2/re -> no distinction made by CCs for reassorted clade; all strains are “3C.2a2”.
* A1b in CC definitions includes mutations 62G and 142G that we assign to 131K.
* A1 is more specific in CC definition with additional 121K mutation

### H1N1pdm

* 6B.2 (6b2) may have been misdefined with a “152V” mutation instead of “152T” and was not showing up in 12y tree.
* 6B.1A.1 (A1) was defined by 183P alone, but 183P is not monophyletic. Kept Nextstrain nucleotide markers.
* 6B.1A.3 (A3) defining substitutions T120A, S183P are not monophyletic. Kept Nextstrain nucleotide markers.
* 6B.1A.4 (A4) defining substitutions included A144E. There is no A144E in the tree, but there is a A141E.
* 6B.1A.5a (A5a) defining substitution should be T185I instead of T185A.
* 6B.1A.6 (A6) defining substitutions are not monophyletic, so keeping Nextstrain nucleotide markers.

### Vic

* Nextstrain clades 1A and 1B become V1A and V1B under CC naming. This means Nextstrain’s V1A goes away.
* V1B (1B) defining substitution 58P is not monophyletic, so keeping Nextstrain nucleotide markers.

### Yam

* Y1’s representative strain, B/Florida/4/2006, gets filtered by TreeTime for diverging too much from the clock rate. To annotate Y1 at the root, I added a “Y1” clade with the reference amino acid at position 1 of SigPep.
* CC definition of clade 3 is much more restrictive than Nextstrain’s.
* CC definitions drop 172Q as a clade.

## Testing

I tested locally with sparse 12y builds for each lineage and pushed these to the staging server for review:

 - [H3N2](https://nextstrain.org/staging/flu/seasonal/h3n2/ha/12y/clades)
 - [H1N1pdm](https://nextstrain.org/staging/flu/seasonal/h1n1pdm/ha/12y/clades)
 - [Vic](https://nextstrain.org/staging/flu/seasonal/vic/ha/12y/clades)
 - [Yam](https://nextstrain.org/staging/flu/seasonal/yam/ha/12y/clades)


rneher:
Thanks, John! This looks very good to me. I have only two comments/questions:

Do you know why K326R is defining for 3C.3a? I would have moved this up one split:
image

3c.2a2/res doesn't appear in Becky's list afaikt

huddlej:
> Do you know why K326R is defining for 3C.3a?

I agree. The defining substitutions for 3C.3a were A138S, F159S, N225D, and K326R. I suspect the 326 was added such that the clade starts at the previous vaccine strain, A/Switzerland/9715293/2013.

> 3c.2a2/res doesn't appear in Becky's list afaikt

That's right; I kept it here because that reassortment event was so important at the time and it seems useful to have a way to reference it. I think in past conversations you and @trvrb have been OK with dropping A2/re, so I can drop it here.

rneher:
I don't have strong feelings regarding `/re`. I do think it would make sense to remove `K326R` from the definition of 3C.3a, but I am ok with keeping it as is if that is what they want. doesn't really matter for any present day virus anyway.

rneher (approved):

PR #75, opened 2021-12-10
Update clade definition H3N2 3C.2a1b.2a

joverlee521:
The 193M mutation does not appear in the H3N2 12y tree for HA1, but it
does appear for HA2.


huddlej (approved):
That looks much better. Thank you!

PR #76, opened 2022-01-12
Refactor workflow

rneher:
This PR refactors the old work flow that was spread out over files like `Snakefile_base`, `Snakefile_WHO`, etc into a structure that follows the recommendations. In particular, it aims at splitting the workflows into substeps that have defined input and output files to minimize the inter-dependencies of different parts of the workflow. Complex input functions are avoided as much as possible to increase legibility.

## TODO

 - [x] Rename new build outputs to match the current Auspice filenames (for example, `flu_seasonal_h3n2_2y_ha.json` from the refactored workflow needs to be `flu_seasonal_h3n2_ha_2y.json` to match the live builds). Some options include:
   - Allow the build config to define an auspice_name that parallels but remains decoupled from the build_name so the build name could be `flu_seasonal_{lineage}_6m` but the auspice_name could be `flu_seasonal_{lineage}_{{segment}}_6m`. There is some precedent for this approach in [the ](https://docs.nextstrain.org/projects/ncov/en/latest/reference/workflow-config-file.html#auspice-json-prefix)auspice_json_prefix[ option in the ncov workflow](https://docs.nextstrain.org/projects/ncov/en/latest/reference/workflow-config-file.html#auspice-json-prefix).
   - Add a custom rule to rename public build outputs (e.g., with `cp`) [like we do for dated JSONs in ncov](https://github.com/nextstrain/ncov/blob/822a77ba070d1d37483d45c22a00b5f83d16e974/workflow/snakemake_rules/export_for_nextstrain.smk#L349).
   - Update on Jan 25, 2023: This is now implemented for the public builds through custom renaming rule plus an `auspice_name` build parameter. To get all of the properly named public builds we just need to run: `snakemake --configfile profiles/nextstrain-public.yaml -- all_public`. This `all_public` target will generate the expected files in the `auspice_renamed/` directory.
 - [x] Root trees with the alignment reference and then remove that root sequence after rooting. Some options include:
   - Implement proposal in https://github.com/nextstrain/augur/issues/340
   - Copy [custom rule to prune outgroup after refine step](https://github.com/nextstrain/seasonal-flu/blob/5beeb5e53e073ef4ef66df17bdb53e76958c6a0c/Snakefile_base#L544) from the workflow on master branch
   - Update on Jan 27, 2023 by @huddlej: I ended up [copying and modifying the prune outgroup script](https://github.com/nextstrain/seasonal-flu/pull/76/commits/6457b6f6e65852218868d5eb700e89ac0a45da50) to prune the outgroup using the same reference FASTA used for alignment. I also [moved the pruning immediately after tree building](https://github.com/nextstrain/seasonal-flu/pull/76/commits/7f2045c967ed7642cac2d2937ed428ad9c008e11) and removed the midpoint rooting logic from the "sanitize" script to avoid undoing the rooting that happens during tree building.
 - [x] Port rules to build Auspice JSONs for private.nextflu.org
 - [x] Verify that public and private builds run and deploy as expected

## Running public builds (40 trees)

```bash
nextstrain build \
  --aws-batch \
  --detach \
  --cpus 36 \
  --memory 72gib \
  . \
  all_public \
  -p \
  --configfile profiles/nextstrain-public.yaml
```

Output is in `auspice_renamed/`.

## Running private.nextflu.org builds (80 trees)

```bash
nextstrain build \
  --aws-batch \
  --detach \
  --cpus 36 \
  --memory 72gib \
  . \
  all_who \
  -p \
  --configfile profiles/private.nextflu.org.yaml
```

Output is in `auspice-who/`.

## Future TODOs

 - [ ] Copy all reference sequences into their respective directories and split them into FASTA and GFF files (using a script to convert GenBank to FASTA + GFF)


workflow/snakemake_rules/export.smk
```diff
@@ -0,0 +1,54 @@
...
+        rules.traits.output.node_data
+    ]
+
+    if config.get('titer-models',False):
+        inputs.append(rules.titers_sub.output.titers_model)
+        inputs.append(rules.titers_tree.output.titers_model)
+    if config.get('glycosilation', False) and wildcards.segment in ['ha', 'na']:
+        inputs.append(rules.glyc.output.glyc)
+    if config.get('lbi', False) and wildcards.segment in ['ha', 'na']:
+        inputs.append(rules.lbi.output.lbi)
```
huddlej, line 22: These top-level boolean keys like `titer-models` don't allow us to limit which segments to apply the models to from the config, so we'll run titer models on NA, PB1, etc. Could we move these types of parameters into the build definitions (like the `vaccines` option below) instead so we have more control over which builds use which annotations?

Then we can eliminate all of these additional segment tests that happen for glycosylation, LBI, etc.

workflow/snakemake_rules/select_strains.smk
```diff
@@ -0,0 +1,129 @@
...
+
+def get_subsample_input(w):
+    files = {"metadata": f"data/{config['builds'][w.build_name]['lineage']}/metadata.tsv"}
+    if config['builds'][w.build_name]['subsamples'][w.subsample].get('priorities', '')=='titers':
+        files['titers']=build_dir + f"/{w.build_name}/titer_priorities.tsv"
+    return files
+
+rule subsample:
+    input:
+        unpack(get_subsample_input)
```
huddlej, line 51: Instead of unpacking output from a wildcard function like this, what if we allow users to define the “titer priorities” file as a subsample input in the config file (e.g., `priorities_file: "{build_name}/titer_priorities.tsv"`)? This would allow Snakemake to build the titer priorities on the fly, as needed, and also allow users to define their own priorities file. This would decouple the *titers* bit from the *priorities* bit.

workflow/snakemake_rules/select_strains.smk
```diff
@@ -0,0 +1,129 @@
...
+        with open(output.strains, 'w') as fh:
+            fh.write('\n'.join(strains)+'\n')
+
+rule select_sequences:
+    input:
+        strains = build_dir + "/{build_name}/strains.txt",
+        sequences = lambda w: f"data/{config['builds'][w.build_name]['lineage']}/{w.segment}.fasta"
+    output:
+        sequences = build_dir + "/{build_name}/{segment}/sequences.fasta"
+    shell:
```
huddlej, line 90: We can do all of this with a single augur filter command. I don’t think our flu data are so big that we need to use seqkit and custom rules to concat strain lists. The `select_sequences` rule should just be a call to augur filter with `--exclude-all --include <strain list files> --output-sequences sequences.fasta --output-metadata metadata.tsv`. It will be fast enough and allow us to drop a bunch of separate rules.

workflow/snakemake_rules/select_strains.smk
```diff
@@ -0,0 +1,129 @@
...
+    output:
+        metadata = build_dir + "/{build_name}/metadata.tsv"
+    run:
+        import pandas as pd
+        with open(input.strains) as fh:
+            strains = [x.strip() for x in fh]
+
+        d = pd.read_csv(input.metadata, sep='\t', index_col=0).loc[strains]
+
+        d.to_csv(output.metadata, sep='\t')
```
huddlej, line 108: Ditto the comment above

huddlej (commented):
Thank you for putting this together, @rneher! This is already a major improvement over the current workflow both in the separation of concerns into separate rules and the abstraction of the build configs into separate config files. I like that we can use the powerful array builds approach when we want, but that the simpler flat config approach works just as well.

My biggest structural questions have to do with who we think will use this workflow:

1. Do we want to support this as a canonical seasonal flu workflow that other public health organizations can run on their own data like we do for SARS-CoV-2?
2. Can we support definition of specific input sequences and metadata in the config, when fauna inputs are not an option (both for our own internal uses and for external users).

In reviewing this and what a user's experience would be like working from GISAID, I noticed that we can now download ~20,000 sequences at once from GISAID. This means most people could run some powerful analyses from a couple of simple GISAID downloads. For an H3N2 HA and NA analysis, for example, this would produce two FASTA files like `gisaid_epiflu_sequence_2015-2018_ha.fasta` and `gisaid_epiflu_sequence_2015-2018_na.fasta` that would need to run through `augur parse` to get the separate sequences and metadata. Can we support this simple use case in this workflow even for our own work?

It would also be nice to eventually move all of the configuration bits out of the Snakefiles and into config files. For example, clock rates, gene names, etc. are all candidates to move into config files where they are more visible, easier for non-Snakemake experts to modify, etc.

Snakefile
```diff
@@ -1,349 +1,27 @@
...
-    params:
-        "results ",
-        "targets ",
-        "auspice ",
-        "auspice-who ",
-        "logs ",
-        "data"
-    shell:
-        "rm -rfv {params}"
+        [f"auspice/{b}_ha.json" for b in config[build_dir + ""]] + \
```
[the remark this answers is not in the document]
rneher: I this should be rethought. this was mostly my quick way of hacking together various targets for testing. But being able to specify an auspice dir has been useful to run different analysis and profile in the same directory and not struggle with clean removing stuff you want to keep.

workflow/snakemake_rules/select_strains.smk
```diff
@@ -0,0 +1,135 @@
+'''
+This file contains rules that select strains for a build, extracts the sequences
+and the meta data subset.
+
+input:
+ - "data/lineage/metadata.tsv"
```
huddlej, line 6: In our discussion today, we want to support users who download separate segment FASTAs from GISAID. We may need to support files like `data/lineage/metadata_segment.tsv`

huddlej:
@joverlee521 made the point that we also want to support example data from open data with an example Auspice JSON so users can confirm that the basic workflow will work for them as a tutorial.

We also need to move the distance files, fitness components, etc. still.

I will try running the workflow with standard GISAID downloads and determine what minimal changes need to happen for this to work.

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -0,0 +1,129 @@
...
+
+rule metadata:
+    input:
+        segment_metadata = lambda w: [f"data/{w.lineage}/metadata_{segment}.tsv" for segment in config['segments']]
+    output:
+        "data/{lineage}/metadata.tsv"
+    params:
+        segments = lambda w: config['segments']
+    run:
+        import pandas as pd
```
corneliusroemer, line 111: It may be worth factoring this out into a script. I've found embededded Python code in rules to be tricky to debug.

huddlej: Agreed; that should be part of our Snakemake style guide. Also, `run` blocks don't work with `conda` environments.

huddlej: [Done](https://wiki.nextstrain.org/t/Snakemake+style+guide). 😄

workflow/snakemake_rules/core.smk
```diff
@@ -0,0 +1,295 @@
...
+ - builds/{build_name}/{segment}/tree.nwk
+ - builds/{build_name}/{segment}/branch-lengths.json
+ - builds/{build_name}/{segment}/aa-muts.json
+ - builds/{build_name}/{segment}/nt-muts.json
+ - builds/{build_name}/{segment}/traits.json
+'''
+
+localrules: clades, sanitize_trees
+
+build_dir = config.get("build_dir", "builds")
```
huddlej, line 21: Using this workflow more, I was surprised by this directory as the place to look for my output. The rest of our workflows use `results/` with the contents of that directory depending on the pathogen workflow (`zika` stores a single set of outputs top-level while `ncov` nests build-specific outputs in subdirectories). Can we switch back to that here?

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -0,0 +1,129 @@
...
+        """
+
+
+rule metadata:
+    input:
+        segment_metadata = lambda w: [f"data/{w.lineage}/metadata_{segment}.tsv" for segment in config['segments']]
+    output:
+        "data/{lineage}/metadata.tsv"
+    params:
+        segments = lambda w: config['segments']
```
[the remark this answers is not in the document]
huddlej: After working with this workflow more, I'm not sure we want a single metadata file per lineage. This approach works when we don’t have any segment-specific metadata, but our work with SARS-CoV-2 suggests we will want this kind of metadata. Some example columns we would want to include are:

* GISAID segment id (the id of the segment sequence, different from the id of the virus isolate)
* GenBank segment id (the accession of the sequence in GenBank)
* Nextclade QC metrics for each sequence (length, number of gaps, runs of SNVs, rare mutations, etc.)
* GISAID- or Nextclade-annotated sequence-specific metadata including amino acid substitutions, nucleotide mutations, insertions, and deletions

These segment-specific columns will get unwieldy especially when we include all segments and end up with columns like `aaSubstitutions_ha`, `aaSubstitutions_na`, `aaSubstitutions_mp`, etc. for every combination of segment and segment-specific column. I think we should keep the pattern of one metadata file to one sequence file with the segment-specific data in the standard "long" format and include a "segment" column that we can query on. We can handle the set logic of which strains have the segments we want using other tools.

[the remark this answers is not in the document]
huddlej: Ok, after posting that comment, I almost immediately ran into the issue that probably motivated the single metadata file; augur filter doesn't work with multiple segment inputs, so we can't subsample the same strains across segments unless we first collect the metadata together into a single file.

This potential multi-segment functionality of augur filter (or joining of multiple metadata files by a key) relates to and potentially conflicts with our desire to support multiple metadata inputs for nonsegmented genomes (or concatenating of multiple metadata files). I'll work on moving the code from the `run` block that handles this logic in this PR into its own function, but maybe the better long-term solution is to support this kind of multi-segment joining as a separate augur command that could be piped into augur filter, instead of expecting augur filter to know about multiple segments.

workflow/snakemake_rules/core.smk
```diff
@@ -0,0 +1,291 @@
...
+    shell:
+        """
+        augur tree \
+            --alignment {input.alignment} \
+            --tree-builder-args {params.tree_builder_args} \
+            --output {output.tree} \
+            --nthreads {threads}
+        """
+
+rule sanitize_trees:
```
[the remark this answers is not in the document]
huddlej: Just a note that the new script for joining metadata (`scripts/join_metadata.py`) supports inner joins of the data, so we could choose to limit analyses to strains with all requested segments at the beginning of the workflow. The steps that follow that join and work from the joined metadata should always pull the same strains for tree building...

huddlej:
I forgot to mention that moving the augur parse/join-metadata logic into `select_strains.smk` allowed me to run builds with FluDB and GISAID data after a little bit of data wrangling. With data from both sources, the workflow crashed at the tree building step because IQ-TREE found duplicate entries for the same sequence names. I added a rule to deduplicate sequences by name using `seqkit rmdup` and this allow the workflow to run for both data sources.

I download a FASTA file from FluDB (with metadata in the deflines) into `data/h3n2/raw_ha.fasta` and defined a config file like so:

```yaml
fasta_fields: "strain accession segment date country state"
prettify_fields: "country state"

lat-longs: "config/lat_longs.tsv"
segments:
  - ha

tree:
  tree-builder-args: "'-ninit 10 -n 4 -czb'"

submission_date_field: date
recency:
  date_bins: [7, 30, 90]
  date_bin_labels: ["last week", "last month", "last quarter"]
  upper_bin_label: older

builds:
    "example_fludb_build":
      reference: "config/h3n2/{segment}/reference.fasta"
      annotation: "config/h3n2/{segment}/genemap.gff"
      tree_exclude_sites: "config/h3n2/{segment}/exclude-sites.txt"
      clades: "config/h3n2/ha/clades.tsv"
      lineage: "h3n2"
      min_date: "2019-01-01"
      auspice_config: "config/h3n2/auspice_config.json"
      subsamples:
        global:
            filters: "--group-by year month --subsample-max-sequences 200 --min-date 2019-01-01"
```

The workflow produced this tree (apparently no one has submitted flu data to flu db since the pandemic??):

image

I download the GISAID sequences with only the strain name in the FASTA to `data/h3n2/ha.fasta`. I also downloaded the XLS spreadsheet of metadata for the same records, manually converted the XLS to XLSX with Excel, and then ran the following transforms with csvtk to get the cleaned metadata:

```
csvtk xlsx2csv h3n2_ha.xlsx \
  | csvtk cut -f Isolate_Id,Isolate_Name,Passage_History,Location,Submitting_Lab,Authors,Originating_Lab,Collection_Date,Submission_Date,Host_Age,Host_Age_Unit,Host_Gender,Is_Vaccinated \
  | csvtk rename -f Isolate_Id,Isolate_Name,Passage_History,Location,Submitting_Lab,Authors,Originating_Lab,Collection_Date,Submission_Date,Host_Age,Host_Age_Unit,Host_Gender,Is_Vaccinated -n gisaid_epi_isl,strain,passage,location,submitting_lab,authors,originating_lab,date,date_submitted,host_age,host_age_unit_,host_sex,is_vaccinated \
  | csvtk --tabs --out-tabs uniq -f strain > data/h3n2/metadata.tsv
```

Then, I defined a GISAID build config like this:

```yaml
fasta_fields: ""
prettify_fields: ""

lat-longs: "config/lat_longs.tsv"
segments:
  - ha

tree:
  tree-builder-args: "'-ninit 10 -n 4 -czb'"

submission_date_field: date_submitted
recency:
  date_bins: [7, 30, 90]
  date_bin_labels: ["last week", "last month", "last quarter"]
  upper_bin_label: older

builds:
    "example_gisaid_build":
      reference: "config/h3n2/{segment}/reference.fasta"
      annotation: "config/h3n2/{segment}/genemap.gff"
      tree_exclude_sites: "config/h3n2/{segment}/exclude-sites.txt"
      clades: "config/h3n2/ha/clades.tsv"
      lineage: "h3n2"
      min_date: "2019-01-01"
      auspice_config: "config/h3n2/auspice_config.json"
      subsamples:
        global:
            filters: "--group-by year month --subsample-max-sequences 200 --min-date 2019-01-01"
```

The workflow produced this tree:

image

The data prep for GISAID above is not ideal because it resolves duplicates by taking the first sequence and metadata record. With a little extra work, we should be able to select the most recent accession for a given strain name as the preferred sequence and metadata record.

profiles/nextstrain-public/rename.smk
```diff
@@ -0,0 +1,47 @@
...
+    output:
+        "auspice_renamed/{auspice_name}.json",
+    shell:
+        """
+        cp {input} {output}
+        """
+
+rule rename_auspice_root_sequence:
+    input:
+
```
joverlee521, line 31: Missing input 👀

profiles/nextstrain-public/rename.smk
```diff
@@ -0,0 +1,47 @@
...
+    return ""
+
+rule rename_auspice_main:
+    input:
+        _get_file_by_auspice_name,
+    output:
+        "auspice_renamed/{auspice_name}.json",
+    shell:
+        """
+        cp {input} {output}
```
joverlee521, line 26: ICYMI, there was a recent [discussion in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1673567129704069) about avoiding using `cp` in workflows. I don't think disk space will be too much of an issue with these JSONs, but switching to `ln` can improve runtime.

profiles/nextstrain-public.yaml
```diff
@@ -173,9 +190,12 @@ array-builds:
         - 2Y
       reference_min_date:
         - 8Y
+      resolution:
+        - 2y
     build_name: "flu_seasonal_h3n2_2y"
     build_params:
       lineage: h3n2
+      auspice_name: "flu_seasonal_{lineage}_{{segment}}_2y"
```
joverlee521, line 198: Should this also be using the `resolution` wildcard? 

```suggestion
      auspice_name: "flu_seasonal_{lineage}_{{segment}}_{resolution}"
```

profiles/nextstrain-public/rename.smk
huddlej: Holy cow! That's an end-of-the-day-emacs error if I ever saw one. Fixed now.

profiles/nextstrain-public/rename.smk
huddlej: Thanks! I had thought about using symlink here, but I forgot about hardlinking. Fixed in 817f197.

profiles/nextstrain-public.yaml
huddlej: The 2y H3N2 build is a special case with its own build params, so this would technically work, but it's clearer to just always use the resolution variable throughout. Fixed in 0a4db33.

PR #85, opened 2022-06-17
Use augur io and dates

joverlee521:
Replaces all uses of `augur.utils.read_metadata` with `augur.io.read_metadata` and uses of `augur.utils.get_numerical_dates` with `augur.dates.get_numerical_dates`. 

The new dates module was introduced in augur 16.0.0 so I also updated the augur dependency in the environment.yaml file.

### Related issue(s)
Related to https://github.com/nextstrain/augur/pull/934 and https://github.com/nextstrain/augur/issues/970

### Testing
Verified that `select_strains` still ran as expected (not sure how to compare outputs since it's a random subsampling)
Checked the following produce the same outputs before and after change:
- age_distributions
- construct-recency-from-submission-date (only difference is augur version)
- full_region_alignments
- generate_locations_file
- scores

### DIFF
--- a/scripts/construct-recency-from-submission-date.py
+++ b/scripts/construct-recency-from-submission-date.py
@@ -1,7 +1,6 @@
 import argparse
 from augur.utils import read_metadata, write_json
 import datetime
-import json
 import numpy as np
 import pandas as pd
 
--- a/scripts/construct-recency-from-submission-date.py
+++ b/scripts/construct-recency-from-submission-date.py
@@ -1,5 +1,6 @@
 import argparse
-from augur.utils import read_metadata, write_json
+from augur.io import read_metadata
+from augur.utils import write_json
 import datetime
 import numpy as np
 import pandas as pd
@@ -53,8 +54,7 @@ if __name__ == '__main__':
     )
     args = parser.parse_args()
 
-    meta, columns = read_metadata(args.metadata)
-    meta = pd.DataFrame.from_dict(meta, "index")
+    meta = read_metadata(args.metadata)
     meta[args.submission_date_field] = pd.to_datetime(meta[args.submission_date_field])
 
     node_data = {'nodes': {}}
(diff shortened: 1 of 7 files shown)


scripts/construct-recency-from-submission-date.py
```diff
@@ -1,7 +1,7 @@
 import argparse
-from augur.utils import read_metadata, write_json
+from augur.io import read_metadata
+from augur.utils import write_json
 import datetime
-import json
```
victorlin, line 4: Small thing, but this would fit better under 14bc83a7972295e4570f4954a8c13ee60df06641 instead of f79489c7e59a2969c7db7da51004b9be8f128a49

scripts/generate_locations_file.py
```diff
@@ -1,7 +1,7 @@
-import numpy as np
```
victorlin, line 1: Small thing, but this would fit better under 14bc83a7972295e4570f4954a8c13ee60df06641 instead of f79489c7e59a2969c7db7da51004b9be8f128a49

scripts/select_strains.py
```diff
@@ -2,12 +2,13 @@
 import sys
 import os
 from collections import defaultdict
-from datetime import datetime, timedelta, date
```
victorlin, line 5: Small thing, but this would fit better under 14bc83a7972295e4570f4954a8c13ee60df06641 instead of f79489c7e59a2969c7db7da51004b9be8f128a49

victorlin (approved):

scripts/construct-recency-from-submission-date.py
joverlee521: Thanks, moved to 06f0b8b in latest push.

scripts/generate_locations_file.py
joverlee521: Thanks, moved to 06f0b8b in latest push.

scripts/select_strains.py
joverlee521: Thanks, moved to 06f0b8b in latest push.

joverlee521:
Successfully ran this week's flu builds on this branch with the `nextstrain/base:latest` image on AWS Batch.

PR #86, opened 2022-06-17
Fix age in months detection

victorlin:

Previously, the condition checked if the age string was exactly 'm'. Updated to check the last character.

### Related issue(s)

_N/A_

### Testing

Not sure how to test.


joverlee521 (approved):
Thanks for catching this @victorlin! LGTM.

Issue #98, opened 2023-02-14
H3N2 NA aa length is wrong 


ammaraziz:
### Current Behavior

The H3N2 NA build on nextstrain.org contains a mutation that's not within the range for selection. There is a major mutation (I469T) that is not selectable by the 'Color By' option. The range is 1 - 468. Note the mutation is outside this range (I469T)

### Expected behavior

This link should work:
https://nextstrain.org/flu/seasonal/h3n2/na/2y?branchLabel=aa&c=gt-NA_469

But nextstrain will return the standard tree. See NODE_0001125 for the mutation I am referring to.

### Possible solution
Change the range to 1 - 469

ammaraziz:
Out of interest is this an auspice setting or an issue with augur output?

joverlee521:
Hi @ammaraziz, thank you for reporting this bug! 

The root of the problem seems to be the [CDS coordinates in our workflow's H3N2 NA reference](https://github.com/nextstrain/seasonal-flu/blob/master/config/reference_h3n2_na.gb#L47). The current coordinates are `4..1410`, but the full coordinates should be `4..1413` based on the [record in GenBank](https://www.ncbi.nlm.nih.gov/nuccore/CY113679.1). 

This reference is passed to the `augur translate` command via the `--reference-sequence` option in the workflow. The [features parsed from the reference](https://github.com/nextstrain/augur/blob/53fed2dcbb8619efbf76766e2b92558e201cf09a/augur/translate.py#L382-L387) are used to generate the genome annotations which are ultimately exported by `augur export v2` as the `meta.genome_annotations` in the main Auspice JSON file. 

Within Auspice, the range displayed is based on the [gene length calculated from the `meta.genomeAnnotations`](https://github.com/nextstrain/auspice/blob/603495ddeb998dcb96d7f25a87abbb0b2166e7de/src/actions/recomputeReduxState.js#L311-L317). With the current CDS coordinates, this is calculated as `Math.floor((1410 - 4) / 3)`, which is 468.

joverlee521:
For now, I've manually edited the genome annotations in the current seasonal flu builds so you can use the color-by option. 
Once #99 is merged, future builds should be fixed as well.

tsibley:
Notably, even in the copy we have with the bad coordinates, the `translation` field for that CDS _is_ the correct length (469).

This also raises the question of if any of the other reference files are subtly wrong?

ammaraziz:
[...]

PR #99, opened 2023-02-15
config/reference_h3n2_na: Update CDS coordinates

joverlee521:

Update CDS coordinates to match the current record in GenBank and to resolve issue #98.

### Testing

- [x] Checks pass
- [x] Will test with the next run of flu builds.

### DIFF
--- a/config/reference_h3n2_na.gb
+++ b/config/reference_h3n2_na.gb
@@ -44,7 +44,7 @@ FEATURES             Location/Qualifiers
                      /db_xref="IRD:NIGSP_CEIRS_CIP047_RFH3_00116.NA"
      gene            4..1413
                      /gene="NA"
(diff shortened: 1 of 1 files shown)


joverlee521:
This was tested with last week's flu builds. The deployed builds currently at https://nextstrain.org/flu/seasonal/h3n2/na/2y have the correct range from 1 to 469 and the coloring for the NA gene at site 469 works as expected now.

joverlee521:
This should not affect the refactor workflow in #76 since the new workflow uses a [GFF file with the correct coordinates](https://github.com/nextstrain/seasonal-flu/blob/09f8334e3d6629d253cccaadb1d2c58696f83012/config/h3n2/na/genemap.gff#L2).

PR #102, opened 2023-04-03
Clean up

huddlej:

Clean up unused rules, config files, and scripts from the previous workflow implementation.

Each commit removes a different set of related files, so if people disagree about any set being removed, we can drop the offending commit. For example, the scripts that I've removed are not currently used, but one could argue that it's valuable to keep the `select_strains.py` script around as a historical document.

I retained a few scripts that are not currently used by the workflow but that appear to be generally useful for the future including:

* `scripts/generate_locations_file.py`
* `scripts/mutation_statistics.py`

### Testing

- [x] Checks pass


joverlee521 (approved):
🎉 

For historical scripts, we can always add a "history" section to the README that has permalinks to the scripts.

PR #103, opened 2023-04-03
Upload center-specific titers

huddlej:

Adds titer collections for each lineage, collaborating center, passage type, and assay type to the S3 upload config. When these files are available through S3, we can switch the build configs for public and private.nextflu.org builds to download from S3 instead of fauna, since those builds currently need center-specific titers.

Note that we may want to remove Yam from the upload config after its initial uploads are complete, since we don't expect any new data for this lineage.

### Testing

- [x] Checks pass
- [x] [Run GitHub Action for upload](https://github.com/nextstrain/seasonal-flu/actions/runs/4600056488) and confirm that new titers get uploaded
- [x] Try running first steps of public and private workflows with S3 as input data source


huddlej:
Thanks, @joverlee521! That's right that you'll have to wait ~40 minutes for the fauna-to-S3 transfer for now, but I love the idea of triggering the new builds with the completion of the S3 upload. Slightly longer term, we could revisit the flow of data from various sources into fauna and see where we could replace fauna uploads with direct S3 uploads.

PR #105, opened 2023-04-23
Feat/treeknit refactor

rneher:
This PR adds a set of rules that replace the refine rule in a way that uses the output of TreeKnit and TreeTime to provide better resolved trees and label MCCs to identify reassortment events. 

There additional steps are currently only used in the Europe profile.

### DIFF
--- a/profiles/europe/builds.yaml
+++ b/profiles/europe/builds.yaml
@@ -1,5 +1,6 @@
 custom_rules:
   - workflow/snakemake_rules/download_from_s3.smk
+  - workflow/snakemake_rules/treeknit.smk
 
 fauna_fasta_fields:
   - strain
--- a/scripts/make-branch-length-json.py
+++ b/scripts/make-branch-length-json.py
@@ -1,0 +1,108 @@
+import argparse
+from treetime.arg import assign_mccs
+from Bio import Phylo
+import numpy as np
+import pandas as pd
+import json
+import random
+
+def to_float(x):
+    try:
+        return float(x)
+    except:
+        return None
+
+def get_clock(fname):
+    with open(fname) as fh:
+        for line in fh:
+            if 'rate:' in line:
+                rate = float(line.strip().split(':')[-1])
+
+    return rate
+
+if __name__=="__main__":
+    parser = argparse.ArgumentParser(
+        description="turn newick trees into branch length jsons",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+
+    parser.add_argument('--timetree', type = str, help='timetree for clock length.', required=True)
+    parser.add_argument('--divtree', type = str, help='divtree for branch length.', required=True)
+    parser.add_argument('--molecular-clock', type = str, help='file with clock rate.', required=True)    
+    parser.add_argument('--mccs', type = str, help='mccs.', required=True)
+    parser.add_argument('--dates', type=str, required=True)
+    parser.add_argument('--output-tree', type=str, required=True)
+    parser.add_argument('--output-node-data', help="name of the file to write node data to", required=True)
+    args = parser.parse_args()
+
+    dates = pd.read_csv(args.dates, sep='\t', index_col=0, skiprows=1)
+    rate = get_clock(args.molecular_clock)
+
+    timetree = Phylo.read(args.timetree, 'nexus')
+    timetree.root.up = None
+    for n in timetree.get_nonterminals():
+        for c in n:
+            c.branch_length *=  rate
+            c.up = n
+
+    divtree = Phylo.read(args.divtree, 'nexus')
+
+    MCCs = []
+    n_large_mccs = 0
+    with open(args.mccs) as fh:
+        for line in fh:
+            if line.strip():
+                MCCs.append(line.strip().split(','))
+                if len(MCCs[-1])>2:
+                    n_large_mccs += 1
+
+
+    MCCs.sort(key=lambda x:len(x), reverse=True)
+    mcc_map = list(range(len(MCCs)))
+    letters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+    random.seed(987)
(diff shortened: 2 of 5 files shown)


huddlej (commented):
I'm sorry I didn't formally review this here after we discussed it in person, @rneher. This can just be a note that we can reference in future discussions.

I like the way that this custom implementation of refine for ARGs does not interfere with the core workflow but can be seamlessly pulled in when needed (e.g. the European builds). I also like the idea of using a custom script in a single repo like this before moving the logic into an Augur subcommand.

We discussed the obvious issue that TreeKnit only supports two segments right now and that we will eventually/soon want builds for all 8 segments. Since HA and NA are the only segments that need regularly updated builds, we could have a separate profile for the other segments that does not use the TreeKnit logic and only needs to be manually run ahead of the TCs and VCMs (so, 6 times a year).

Another more pressing issue with using TreeKnit in this workflow is the difficulty of setting up a Bioconda recipe for TreeKnit. [I've previously attempted this recipe](https://github.com/bioconda/bioconda-recipes/pull/34130) and failed to get the Julia build environment to produce a working binary. Pierre didn't seem optimistic about this situation, but I wonder whether a @corneliusroemer, @ivan-aksamentov, or other Bioconda wizard could help push that recipe to completion, so everyone could use TreeKnit without manually installing it outside of the Conda environment.

tsibley:
It might be acceptable for the Bioconda recipe to distribute pre-built binaries instead? IIRC that's what's being done for Nextclade/Nextalign. Presumably portable, statically-linked pre-built versions TreeKnit are already available?

huddlej:
That would be a great solution, @tsibley! I don't think those statically-linked binaries exist yet, but it seems that [the Julia PackageCompiler tool](https://julialang.github.io/PackageCompiler.jl/stable/) should be able to produce these. [TreeKnit installation](https://pierrebarrat.github.io/TreeKnit.jl/) depends on installing through Julia currently. [This commit was my attempt to build binaries with PackageCompiler](https://github.com/bioconda/bioconda-recipes/pull/34130/commits/a7d1ef0d680ba8ca07c13a8309594ddf10ca5bdd), but I couldn't get the Julia build environment setup properly.

tsibley:
Ahh. Hmm. Were you able to build locally? That is, were the build issues due to the peculiarities of the BioConda CI environment or more general? If we can build it outside of BioConda, then that should be enough.

ivan-aksamentov:
@tsibley I tried, spent an evening (each build attempt takes 15-30 min) and still don't have any idea. I heard Julia language is cool for scientific exploration, but if you try to actually ship something written in it, it turns out that the ecosystem/community lives in a parallel universe, orthogonal to everything known in software engineering so far, and they are currently figuring out: ["Applications: what are they?" #1962](https://github.com/JuliaLang/Pkg.jl/issues/1962). They are not in a particular hurry either.

If anyone will be trying, then here is my current understanding:
 - Julia is a REPL-first ecosystem. There are [ways](https://pierrebarrat.github.io/TreeKnit.jl/) to build a distribution of a sort (it's called "to build" or "to install"). But all these produce a non-portable mess, with hardcoded paths, like `/home/username/...` etc. in your `~/.julia` directory. Sort of like `pip install .`, but worse. Not even mentioning building for different OS, architecures and libc. These will be a hell. Let's focus on building on one platform.
 - Pierre is using [Comonicon package](https://comonicon.org/stable/) to generate a CLI [here](https://github.com/PierreBarrat/TreeKnit.jl/blob/abf9c8c2050e054534f5dd16e0b64aea2e479d3b/src/cli.jl#L49). It's something similar to [docopt](https://github.com/docopt/docopt). Comonicon [should generate](https://github.com/PierreBarrat/TreeKnit.jl/blob/3e5a4b8d6edb158f95370cd72467bda318f82f4d/deps/build.jl#L2) an (invisible to us) entry point function `TreeKnit.command_main()`
 - We should be able to refer to this function when compiling. My best attempt was (Julia code):
   ```julia
   PackageCompiler.create_app(".", "build_dir/", executables=["TreeKnit" => "TreeKnit.command_main"])
   ```
   but this ended with absolutely non-informative error (Did not record it and a repro will take half an hour). Note that you can also put a name of a non-existent function and it will haply compile something for the best part of an evening, succeed, and then when you execute the result it will tell that there is no such function. Real time saver!
 - Note that last time I tried with another project, we needed to [trace](https://github.com/neherlab/pangraph/blob/master/trace.jl) the execution of functions, by running the code, using some input data. And then julia would record the LLVM instructions of that run or something like that, and produce a semblance of a distribution. Except it was non portable, so we simply made a docker container.
 - Comonicon [recommends](https://comonicon.org/stable/project/) to build the app like that (bash code):
   ```bash
   julia --project deps/build.jl app
   ```
   however I am getting "unknown command app" error. I believe that it might require tracing to be configured in Comonicon.toml.
 - However, I believe that `PackageCompiler` might be able to figure things out without tracing nowadays.
 - The most useful insight into how much Julia is messed up I found on [home page of PackageCompiler docs](https://julialang.github.io/PackageCompiler.jl/stable/), the "There are three main workflows" paragraph.
 - Note that in any case you'd need to produce the sysimage (`sys.so`), and an executable. Plus there will be a bunch of other libs on which julia depends. So it is definitely nothing like a standalone executable.


Now, there is a hypothesis that reimplementing the algorithm in an actual programming language  might be easier then struggling with this new game-changing compiler technology.

huddlej:
> Were you able to build locally?

@tsibley I don't think I ever tried a Conda local build, but it seems like the clear next step to rule out CI environment issues.

> Now, there is a hypothesis that reimplementing the algorithm in an actual programming language might be easier then struggling with this new game-changing compiler technology.

This was Pierre's suggestion when we discussed in person last month.

In any case, thank you for trying this out, @ivan-aksamentov!

tsibley:
Wow.

Alright then.

PR #106, opened 2023-05-04
Automate builds

joverlee521:

Add option to trigger builds after the uploads to S3 have completed.
The downstream triggers are optional, so we can still run upload jobs without starting builds. 
I kept the builds as separate GH workflows so that they can also be run independently of the upload job. 

There's a lot of overlap in the public and private build workflows (as well as with all the other AWS Batch workflows across all of our pathogen repos). Hopefully will be able to trim them down after we set up a reusable workflow for these jobs. 

Still need to manually download the builds from AWS Batch and deploy them. I haven't set up any auto-deployments or Slack messages (yet). 

### Testing

- [x] Checks pass
- [x] Worked well for today's builds. [upload](https://github.com/nextstrain/seasonal-flu/actions/runs/4886287107) triggered the downstream [public build](https://github.com/nextstrain/seasonal-flu/actions/runs/4886621659) and [private build](https://github.com/nextstrain/seasonal-flu/actions/runs/4886621735).

### DIFF
--- a/.github/workflows/run-private-nextflu-builds.yaml
+++ b/.github/workflows/run-private-nextflu-builds.yaml
@@ -1,6 +1,12 @@
 name: Run the private Nextflu builds
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      dockerImage:
+        description: "Specific container image to use for build (will override the default of `nextstrain build`)"
+        required: false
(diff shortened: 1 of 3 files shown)


joverlee521:
Merging since this worked well for this week's builds. 
There are definitely more improvements to come!

PR #119, opened 2023-08-29
Use new augur ancestral interface

huddlej:

Updates augur ancestral rule in the core workflow to generate ancestral translations in addition to ancestral nucleotide sequences. Removes the separate rule for translation and the corresponding custom script. Updates all downstream rules to reference a new single `muts.json` file. Rules that depended on the translations with internal node sequences do not need to change, since the output files from the updated ancestral rule maintain the same name.

## Testing

- [x] Tested with nextflu-private builds
- [ ] Checks pass


rneher (approved):
nice! this looks great to me!

PR #124, opened 2023-11-08
Nextclade

rneher:
Adds a workflow to produce nextclade data sets


nextclade/Snakefile
```diff
@@ -0,0 +1,375 @@
...
+        augur tree \
+            --alignment {input.alignment} \
+            --output {output.tree} \
+            --nthreads {threads} \
+            --tree-builder-args '-ninit 10 -n 4 -czb' \
+            > /dev/null
+        """
+
+# root using dates in treetime, use 1500 as sequence length (good enough, doesn't matter)
+rule root:
```
huddlej, line 157: I find myself needing this kind of functionality for other projects, too, where I want to root the tree before refining it and often when I don't want a time tree. This is usually in the context of rooting on the same strain I used as the reference for alignment and I usually want to prune that strain after rooting, if it is an older strain. I wrote [this root and prune script](https://github.com/blab/cartography/blob/master/notebooks/scripts/root_and_prune_tree.py) to accomplish this effect for a different project.

But, this all comes back to @trvrb's [issue about providing an option to root and prune the outgroup in Augur](https://github.com/nextstrain/augur/issues/340) which we should probably just implement soon...

nextclade/Snakefile
```diff
@@ -0,0 +1,375 @@
...
+        """
+
+rule parse:
+    input:
+        sequences="data/{lineage}/raw_{segment}.fasta",
+    output:
+        metadata="data/{lineage}_{segment}_metadata.tsv",
+        sequences="data/{lineage}_{segment}_sequences.fasta",
+    params:
+        fields="strainName virus segment EPI_ISL date submission_date region country division location passage collecting_lab submitting_lab age sex",
```
huddlej, line 71: We already have the parsed sequences and metadata in the same S3 bucket as the raw (unparsed) sequences, but do you need to run `augur parse` here to set the field names in the metadata to names that Nextclade's interface expects?

nextclade/Snakefile
```diff
@@ -0,0 +1,375 @@
...
+            --sequences {input.sequences} \
+            --fields {params.fields} \
+            --output-metadata {output.metadata} \
+            --output-sequences {output.sequences}
+        """
+
+
+def genes(w):
+    if w.segment=='ha': return ["SigPep", "HA1", "HA2"]
+    if w.segment=='na': return ["NA"]
```
huddlej, line 84: I noticed the `segment` wildcard constraints at the top of the file included other genes. Do you plan to support those other genes in Nextclade, too? It could be pretty useful for people. If so, maybe we should move this `genes` function to the top of the file next to the wildcard constraints and add the other segments to this function?

nextclade/config/config_dict.yaml
```diff
@@ -0,0 +1,92 @@
...
+    ha:
+      clade_systems:
+        clade:
+          url: "seasonal_A-H1N1pdm_HA/main/.auto-generated/clades.tsv"
+        subclade:
+          url: "seasonal_A-H1N1pdm_HA/main/.auto-generated/subclades.tsv"
+          key: "subclade"
+      refs:
+        CY121680:
+          filter: "--min-date 2014 --probabilistic-sampling  --exclude-where country='south_korea' --group-by year --min-length 1500  --subsample-max-sequences 1500"
```
huddlej, line 14: Why is South Korea excluded?

nextclade/Snakefile
```diff
@@ -0,0 +1,375 @@
...
+    output:
+        "data/{clade}_{lineage}_{segment}_{reference}.tsv",
+    params:
+        offset=lambda w: config["builds"][w.lineage][w.segment]['refs'][w.reference][
+            "clade_offset"
+        ],
+    shell:
+        """
+        perl -F'\\t' -ne \
+            '$F[2]+={params.offset} if $F[1] =~ "nuc"; \
```
huddlej, line 58: I could see this clade offset logic here and in the config files being the place where bugs get introduced. Since we define clades based on coordinates from a specific reference and we already autogenerate the clades YAML and TSV files in the nomenclature repo, would it be crazy to also autogenerate reference-specific clades files in that same process so you'd always know that the clades file you grab belongs to a specific reference and doesn't need to be modified?

nextclade/Snakefile
```diff
@@ -0,0 +1,375 @@
...
+        """
+        AUGUR_RECURSION_LIMIT=10000 \
+        augur export v2 \
+            --tree {input.tree} \
+            --metadata {input.metadata} \
+            --node-data {input.node_data}\
+            --auspice-config {input.auspice_config} \
+            --color-by-metadata {params.fields} \
+            --minify-json \
+            --title "Nextclade reference tree for Influenza type:{wildcards.lineage} segment:{wildcards.segment} with root {wildcards.reference} built on {params.date}" \
```
huddlej, line 312: I like this pattern of including the date in the title! We should use this in the main workflow, too...

nextclade/config/auspice_config.json
```diff
@@ -0,0 +1,50 @@
+{
+  "title": "Genomic epidemiology of Influenza",
+  "build_url": "https://github.com/neherlab/nextclade_data_workflows",
```
huddlej, line 3: Should we update this build URL to reference this repo instead?

huddlej (commented):
I left a few comments on the code below, but I can't test the workflow without `nextclade_v3`. What's the best place to get this?

nextclade/config/auspice_config.json
rneher: yes, updated!

nextclade/config/config_dict.yaml
rneher: I added a comment on this. There is a batch of south korean sequences with messed 5' UTR .

nextclade/Snakefile
rneher: yes, that is the plan. I moved it up and refactored it.

nextclade/Snakefile
```diff
@@ -0,0 +1,375 @@
...
+        node_data = get_node_data,
+        auspice_config= "build/{lineage}/{segment}/{reference}/auspice_config.json"
+    output:
+        auspice_json="build/{lineage}/{segment}/{reference}/auspice.json",
+    params:
+        fields="region strainName country date EPI_ISL",
+        date=datetime.datetime.utcnow().strftime("%Y-%m-%d"),
+    shell:
+        """
+        AUGUR_RECURSION_LIMIT=10000 \
```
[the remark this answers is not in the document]
rneher: removed

rneher:
we'll release an alpha version in 10 days or so. until then, you can either build it or get if from the CI artefacts. Tom linked to them in slack.

nextclade/Snakefile
```diff
@@ -0,0 +1,451 @@
...
+
+rule align:
+    input:
+        sequences="build/{lineage}/{segment}/{reference}/subsample.fasta",
+        annotation="dataset_config/{lineage}/{segment}/{reference}/annotation.gff",
+        reference="dataset_config/{lineage}/{segment}/{reference}/reference.fasta",
+    output:
+        alignment="build/{lineage}/{segment}/{reference}/align.aligned.fasta"
+    params:
+        outdir=lambda w: f"build/{w.lineage}/{w.segment}/{w.reference}/aligned.gene.{{gene}}.fasta",
```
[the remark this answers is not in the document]
huddlej: Fixed in https://github.com/nextstrain/seasonal-flu/pull/124/commits/c8a068433def78b06b7e5d9abd76cd9dc20b34d4 but revert if I'm wrong here.

nextclade/Snakefile
```diff
@@ -0,0 +1,451 @@
...
+    input:
+        tree = rules.root.output.tree
+    output:
+        tree="build/{lineage}/{segment}/{reference}/tree_rooted_pruned.nwk"
+    params:
+        outliers = "build/{lineage}/{segment}/{reference}/tt_out/outliers.tsv"
+    run:
+        import pandas as pd
+        from Bio import Phylo
+        outliers = pd.read_csv(params.outliers, sep='\t', index_col=0)
```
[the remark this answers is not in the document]
huddlej: Fixed in https://github.com/nextstrain/seasonal-flu/pull/124/commits/17eaf0b3e2da2278246b011bb8f1a06ac871b3e1

nextclade/Snakefile
```diff
@@ -0,0 +1,451 @@
...
+    year="\d\d\d\d",
+    lineage=r"h3n2|h1n1pdm|vic|yam",
+    segment = r'pb2|pb1|pa|ha|np|na|mp|ns',
+    reference="[^_/]+",
+
+
+def genes(w):
+    return {
+        'ha': ["SigPep", "HA1", "HA2"],
+        'na': ["N"]
```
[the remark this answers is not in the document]
huddlej: Fixed in https://github.com/nextstrain/seasonal-flu/pull/124/commits/c8a068433def78b06b7e5d9abd76cd9dc20b34d4

huddlej:
@rneher The workflow runs for me now as expected, with the few changes I made above. I tested one of the datasets generated by the workflow (`datasets/h1n1pdm/ha/CY121680`) with [one of the Nextclade UI prototypes](https://nextclade-git-ui-3-nextstrain.vercel.app) and got the following error about parsing the pathogen JSON:

image

We could merge this PR and keep working on it from `master` or we could pull the main workflow changes out of this PR into `master` and keep working here. Eventually people might start depending on the `nextclade` workflow to work on `master`, but I doubt anyone will right now since it's new.

rneher:
thanks John. I think this error you got should not happen with the most recent version. I merged this now. It is a separate workflow and it we can keep improving it on master .

PR #138, opened 2023-12-28
Add Auspice JSON colorings for HA/NA accessions

### LINKED ISSUE #137: Export GISAID accessions from metadata into Auspice JSONs
## Description

Since the same strain can appear multiple times in GISAID, we should provide the unique GISAID accessions we use for each sequence that appears in our trees. We also modify the original strain names, in some cases, to standardize leading zero padding in ids, etc. such that users cannot find records in GISAID's search interface without the accession.

huddlej:

Adds coloring entries for the HA and NA accessions associated with each sequence in a build, so this information appears in the public trees and allows users to determine the sequences that correspond with each strain name.

## Related issue(s)

Closes #137

## Checklist

- [x] Checks pass

### DIFF
--- a/config/h1n1pdm/auspice_config.json
+++ b/config/h1n1pdm/auspice_config.json
@@ -88,6 +88,16 @@
       "type": "categorical"
     },
     {
+      "key": "accession_ha",
+      "title": "Accession (HA)",
+      "type": "categorical"
+    },
+    {
+      "key": "accession_na",
+      "title": "Accession (NA)",
+      "type": "categorical"
+    },
+    {
       "key": "submitting_lab",
       "title": "Submitting lab",
       "type": "categorical"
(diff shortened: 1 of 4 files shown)


j23414:
Running a quick test: https://github.com/nextstrain/seasonal-flu/actions/runs/7350669867

j23414 (approved):
Test run succeeded, and I see that the accessions are being included as expected in the downloaded Nexus file (and not in the nwk file). 

[non-blocking] I slightly lean towards having these in the "acknowledgements.tsv" file, but I imagine that using Nexus might be the fastest solution for now. Just a note that there is a slight pause in the site if the user selects "Color By: accession..." from the drop down, but I hope this occurrence will be infrequent and non-blocking.

huddlej:
Thanks, @j23414! The implementation as a coloring is not ideal, but hopefully this will be useful for people and we'll eventually setup an augur/auspice interface that doesn't require these to be colorings.

PR #150, opened 2024-02-16
Add measurements panel for Welsh et al. escape scores

huddlej:

Add measurements panel to H3N2 HA build for Welsh et al. escape scores grouped by serum or cohort. Enables interactive visualization and filtering of data.

As we've found with titer data, this view of the raw data allows us to discover patterns that we would miss from summary statistics alone. The following image shows an example of a multimodal distribution for one sample that appears as high variance in the summary view.

image

## Checklist

- [x] Checks pass
- [x] Tested with real build


huddlej:
@genehack This was an idea for an approach that I ended up not taking, so I'll close it. Thanks!

Issue #155, opened 2024-02-27
"Suffolk" in `lat_longs.tsv` points to somewhere unexpected 


fanninpm:
### Current Behavior
In the `lat_longs.tsv` file, Suffolk points to somewhere in Boston Harbor.

### Expected behavior
"Suffolk" should point to somewhere in the East of England/East Anglia, not too far from "Norfolk", "Essex", and "Cambridgeshire".

### How to reproduce
Steps to reproduce the current behavior:
1. Visit https://www.gps-coordinates.net/
2. Copy/paste the coordinates associated with "Suffolk" (which are currently 42.3544455 degrees in latitude and -70.9788771 in longitude)
3. Observe the location on the map

### Possible solution
I don't know enough about how Nextstrain divides the UK in its `lat_longs.tsv` file to give directions for a possible solution. However, I think this may indicate a broader issue as for the structure of the `lat_longs.tsv` file itself. Perhaps there should be more hierarchy in that file, which would answer additional questions such as "Should Hawaii's `region` value be 'Oceania' (because it's out in the middle of the Pacific) or 'North America' (because it's part of the USA)?".

### Additional context
For my purposes, I will use "Suffolk ENG" as my `location` value in my metadata.

joverlee521:
Thank you for reporting this issue @fanninpm!

Looks like the current [Suffolk lat/long](https://github.com/nextstrain/seasonal-flu/blob/45bf4336d9485c1c9bfc44b09b384595d7685032/config/lat_longs.tsv#L10033) points to the same coordinates as the [Suffolk County MA lat/long](https://github.com/nextstrain/seasonal-flu/blob/45bf4336d9485c1c9bfc44b09b384595d7685032/config/lat_longs.tsv#L10035). 

Looking at the commit history (https://github.com/nextstrain/seasonal-flu/commit/90cbd774e92814a1c0aba07b48d87de7f0f5f04a), the seasonal-flu lat_long.tsv was copied over from the [ncov/defaults/lat_longs.tsv](https://github.com/nextstrain/ncov/blob/0af26ae2ec4368251eb9caeadb8ee479e54862d6/defaults/lat_longs.tsv) so we'll need to fix both entries.

PR #156, opened 2024-02-27
Update Suffolk lat/longs

joverlee521:
Fixes https://github.com/nextstrain/seasonal-flu/issues/155

Interestingly, we don't have any entries for "Suffolk" in the fauna/source-data/geo_synonyms.tsv¹ and I couldn't find any metadata in our private S3 bucket that includes "Suffolk" as a location.

¹ https://github.com/nextstrain/fauna/blob/75e309e9afe2cc97fb56d7a24e22a07b698805bf/source-data/geo_synonyms.tsv

## Checklist

- [ ] Checks pass


j23414 (approved):
Looks like the old location is Suffolk VA, while the new one is Suffolk UK. 

I'm good with updating the lat/long if we don't have Suffolk, VA in our dataset elsewhere.

genehack (approved):

Issue #130, opened 2023-11-27
Automate monthly reports


huddlej:
## Context

We produce monthly reports describing seasonal influenza circulation patterns. A substantial portion of time spent on these reports is tedious copying and pasting or running commands manually that could be automated.

## Description

The following specific tasks will minimize the manual work associated with reports and allow us to focus on the science.

- [x] **Schedule nextflu-private builds to run and deploy after data upload (like the public builds)**. This should be easy to do now. We need to add rules that produce dated Auspice JSONs and deploy these JSONs to the desired group.
  - [x] [Add GitHub Action workflow to run nextflu-private builds](https://github.com/nextstrain/seasonal-flu/actions/workflows/run-nextflu-private-builds.yaml)
  - [x] #139
  - [x] #157
- [x] #162
- [x] #174
- [x] #164
- [x] **Automatically generate Markdown tables of new sequences per clade and clade coverage by titer references**. The workflow logic to do this was added in #129, but we still need to add explicit dependencies for [tabulate](https://pypi.org/project/tabulate/) to our Docker and Conda environments.
  - [x] Summarize all new sequences per clade using Nextclade annotations from the full dataset instead of using subsampled data associated with representative or titers builds. Depends on the addition of an earlier step in the workflow that runs Nextclade on all data as mentioned in https://github.com/nextstrain/seasonal-flu/issues/144.
  - [x] #173
- [x] #163
- **Automatically add links for the latest builds to the nextflu-private group landing page after builds have been deployed**.
  - JohnH: I'm flagging this item as unplanned, since I still like to have manual control over updates to the landing page.

genehack:
@huddlej maybe when we meet next week we can discuss what needs to be done to get this particular issue wrapped up?

huddlej:
Closing this now that we've accomplished the major items listed above, but there's always more work to do here in the future! 📈

PR #157, opened 2024-03-14
Trigger nextflu-private builds from upload action

huddlej:

Adds an option to the "upload" GitHub Action to trigger nextflu-private group builds when the upload action completes successfully.

## Related issue(s)

Related to #130 

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/upload.yaml
+++ b/.github/workflows/upload.yaml
@@ -20,7 +20,15 @@ on:
         description: "Specific container image to use for private Nextflu builds"
         required: false
         type: string
-
+      triggerNextfluPrivate:
+        description: "Trigger nextflu-private group builds"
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Looks good to me! 

Is the plan to still only run the nextflu-private group builds before CDC calls? Or should we trigger it with the upload every week?

huddlej:
@joverlee521 I'd trigger it every week, so we can track what's changing on a finer scale than once per month. When we get closer to [automated draft reports and updates to the group landing page](https://github.com/nextstrain/seasonal-flu/issues/130), I would hope weekly builds could be useful for the collaborating center teams outside of our normal meetings.

Issue #144, opened 2024-01-30
Filter low-quality sequences by Nextclade QC status


huddlej:
## Context

Despite our myriad quality filters in the workflow including a global clock rate filter, local clock rate filter, an outliers list, and dropping sequences with poor alignments, our trees still occasionally include low-quality sequences that would have been flagged with a "bad" QC status by Nextclade. In most of my recent experiences with this issue, the low-quality sequences have too many private mutations.

In the best case, these low-quality sequences look strange in the tree. In the worst case, these sequences break the date inference for internal nodes and produce an invalid time tree topology, requiring the builds to be run again.

## Description

Since we already plan to migrate away from Nextalign to Nextclade, we should align sequences with Nextclade and produce the metadata output file with QC statuses. We should filter any sequences that have a QC status of "bad".

We could approach this functionality in a couple of ways:

1. Modify [the existing align rule](https://github.com/nextstrain/seasonal-flu/blob/d1b3a2ec6c1596e9f7c400b7f17fa41d8ab21d72/workflow/snakemake_rules/core.smk#L44) to use Nextclade and produce QC output, add a subsequent post-alignment filter rule before [the tree building rule](https://github.com/nextstrain/seasonal-flu/blob/d1b3a2ec6c1596e9f7c400b7f17fa41d8ab21d72/workflow/snakemake_rules/core.smk#L96) to omit "bad" QC records (using `augur filter` on the alignment sequences and "metadata" from Nextclade), and pass the filtered alignment to the tree rule. We'd probably want to merge the original metadata records with the Nextclade metadata prior to filtering much like [the merge we do in the flu_frequencies workflow](https://github.com/nextstrain/flu_frequencies/blob/9c054762fb265673672c2f6f47b77f7e81db7164/Snakefile#L128-L145).

OR

2. Run Nextclade on all sequences upstream of the main phylogenetic workflow, merge the complete metadata with the Nextclade annotations, upload these combined metadata to S3, start the phylogenetic workflow from the combined metadata files, and apply custom filters on QC in the subsampling logic for each build. The only changes to the main phylogenetic workflow required by this approach would be additions to the build YAML files to include a filter for Nextclade QC status. The bigger changes happen outside of the main workflow in our sequence upload logic.

[...]

PR #158, opened 2024-03-25
Add GitHub Action for Nextclade annotations

huddlej:

Adds rules, config, and GitHub Action file to support running Nextclade on all available HA and NA sequences for H1N1pdm, H3N2, and Vic.

The new Snakemake logic lives in a custom build config named `nextclade` which skips most of the standard workflow build logic, using only the "download from S3" rules and its own custom rules to run Nextclade for the lineages and segments defined in the build config. This custom logic runs Nextclade with the default dataset per lineage and segment [just like the flu_frequencies workflow](https://github.com/nextstrain/flu_frequencies/blob/6e4298fac3361f4a6751d85bcb963064dbb9eee1/Snakefile#L89-L960) and uploads the results to S3. Once we have merged this PR, we should be able to automatically run Nextclade with each ingest of new data and run flu_frequencies from the resulting files on S3.

This PR includes a couple of minor changes to other parts of the standard workflow to allow the Nextclade build config YAML to be as simple as possible and also to allow all workflows to download the parsed sequences and metadata from S3 instead of downloading the raw sequences and parsing them again locally.

To run the Nextclade workflow, use the following command:

```
nextstrain build . upload_all_nextclade_files --configfile profiles/nextclade.yaml
```

This uploads Nextclade annotations and alignment files to S3 per lineage and segment like `seasonal-flu/vic/ha/nextclade.tsv.xz` and `seasonal-flu/vic/ha/aligned.fasta.xz`.

This PR does not produce a merged metadata and Nextclade annotations file like [the one used in the flu_frequencies workflow](https://github.com/nextstrain/flu_frequencies/blob/6e4298fac3361f4a6751d85bcb963064dbb9eee1/Snakefile#L113-L130) or the analogous ncov metadata files with Nextclade annotations included. I stopped short of creating this merged file in S3, too, because we use a single metadata TSV per lineage (for all segments) but we produce Nextclade annotations per segment. We could upload the metadata TSVs per lineage and segment with Nextclade annotations merged into a single file, but this would duplicate a lot of information across segments. Maybe that duplication is acceptable, but it's worth discussing more internally first.

~~Note: since this PR adds a new GitHub Action, I can't test the action until we've merged the PR into master.~~

## Related issue(s)

Related to #144 

## Checklist

- [x] Checks pass
- [x] GitHub Action runs and deploys data to S3 as expected

### DIFF
--- a/.github/workflows/run-nextclade.yaml
+++ b/.github/workflows/run-nextclade.yaml
@@ -1,7 +1,6 @@
 name: Run Nextclade on all sequences
 
 on:
-  pull_request:
   workflow_dispatch:
     inputs:
       dockerImage:
--- a/.github/workflows/run-nextclade.yaml
+++ b/.github/workflows/run-nextclade.yaml
@@ -1,0 +1,32 @@
+name: Run Nextclade on all sequences
+
+on:
+  workflow_dispatch:
+    inputs:
+      dockerImage:
+        description: "Specific container image to use for build (will override the default of `nextstrain build`)"
+        required: false
+        type: string
+
+jobs:
+  run-build:
+    permissions:
+      id-token: write
+    uses: nextstrain/.github/.github/workflows/pathogen-repo-build.yaml@master
+    secrets: inherit
+    with:
+      runtime: aws-batch
+      env: |
+        NEXTSTRAIN_DOCKER_IMAGE: ${{ inputs.dockerImage }}
+      run: |
+        nextstrain build \
+          --detach \
+          --cpus 36 \
+          --memory 72gib \
+          --env AWS_ACCESS_KEY_ID \
+          --env AWS_SECRET_ACCESS_KEY \
+          . \
+          upload_all_nextclade_files \
+          -p \
+          --configfile profiles/nextclade.yaml \
+          --set-threads run_nextclade=16
--- a/.github/workflows/run-nextclade.yaml
+++ b/.github/workflows/run-nextclade.yaml
@@ -29,4 +29,4 @@ jobs:
           upload_all_nextclade_files \
           -p \
           --configfile profiles/nextclade.yaml \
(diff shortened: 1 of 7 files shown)


joverlee521:
Not doing an in-depth review, just reading over this PR because I'm interested in how you handled the different segments. 

> I stopped short of creating this merged file in S3, too, because we use a single metadata TSV per lineage (for all segments) but we produce Nextclade annotations per segment. We could upload the metadata TSVs per lineage and segment with Nextclade annotations merged into a single file, but this would duplicate a lot of information across segments. Maybe that duplication is acceptable, but it's worth discussing more internally first.

It'd be good to discuss what's the easiest for downstream users here (granted that may currently only be the flu-frequencies workflow?). 

> Note: since this PR adds a new GitHub Action, I can't test the action until we've merged the PR into master.

It's possible to test a new GitHub Action with either [Tom's or Victor's work-around](https://github.com/nextstrain/docs.nextstrain.org/commit/aa6f75e103b83676345e3a0f7ac60291aa99e4da)

huddlej:
> It'd be good to discuss what's the easiest for downstream users here

Totally agree, @joverlee521! @plsteinberg is one such downstream user who has to manually join the metadata and HA Nextclade annotations for their project, so we might have an idea of how that could be improved based on their feedback. 😄 

> It's possible to test a new GitHub Action with either [Tom's or Victor's work-around](https://github.com/nextstrain/docs.nextstrain.org/commit/aa6f75e103b83676345e3a0f7ac60291aa99e4da)

Gah, I knew this existed but forgot the mechanics. Thank you for the reminder! Trying with Victor's approach now.

huddlej:
I'm going to merge this now, so I can start running the GitHub Action after our weekly ingests. We should continue to discuss this implementation in the future, though, since there is likely still remove to improve the user experience.

PR #166, opened 2024-05-30
inline root sequence

rneher:


huddlej (approved):
This looks great. Thanks, @rneher!

Issue #178, opened 2024-09-05
Add GISAID isolate ID to acknowledgments download


joverlee521:
Flagged [internally](https://support.nextstrain.org/agent/nextstrain/nextstrain/tickets/details/668036000007764001) by @rneher that the acknowledgements download does not include the GISAID isolate IDs. 

https://github.com/nextstrain/seasonal-flu/pull/138 added the HA/NA _segment_ accessions, but we do not include the isolate ID used to link multiple segments to the same isolate. The isolate id was added to the metadata upload in https://github.com/nextstrain/seasonal-flu/pull/160, so it should be available in the metadata for the builds. 

The fields included in the acknowledgements is [hardcoded in Auspice](https://github.com/nextstrain/auspice/blob/dae78e74d5e06303d901387adf9fe7550a7d4aa9/src/components/download/helperFunctions.js#L326) so the field will have to be named `gisaid_epi_isl`. 

## TODOs
- [x] https://github.com/nextstrain/seasonal-flu/pull/179
- [x] Re-run upload workflow to update S3 column names
- [x] https://github.com/nextstrain/seasonal-flu/pull/180

joverlee521:
@huddlej, would it break any of your workflows to rename the [fasta_fields](https://github.com/nextstrain/seasonal-flu/blob/1677cc2e1a5b824824d2aeb770fe3939dab0d694/profiles/upload.yaml#L28) to `gisaid_epi_isl`?

joverlee521:
Ah, at least the [nextclade workflow](https://github.com/nextstrain/seasonal-flu/blob/1677cc2e1a5b824824d2aeb770fe3939dab0d694/nextclade/Snakefile#L414) is using the `isolate_id` field. 

Maybe better to just rename `isolate_id` to `gisaid_epi_isl` within the phylogenetic workflow right before the `augur export`?

huddlej:
I don't depend on `isolate_id` as a field name anywhere. Since Auspice hardcodes separate fields for GISAID and GenBank accessions (without an awareness of segmented viruses), it makes sense to rename `isolate_id` to `gisaid_epi_isl` in the uploaded metadata.

Unless the Nextclade somehow depends on an `isolate_id` field downstream, I would just change the Nextclade workflow to use the new field name, too.

joverlee521:
Reopening because we've removed `gisaid_epi_isl` from metadata in https://github.com/nextstrain/seasonal-flu/pull/188. 

[...]

PR #179, opened 2024-09-10
Rename metadata column from `isolate_id` to `gisaid_epi_isl`

joverlee521:

The fields to export in acknowledgements is hardcoded in Auspice¹ so the
metadata field needs to be named `gisaid_epi_isl` to be included in
the acknowledgements for the seasonal flu builds. Renaming the column
in our upload config to avoid having to rename it as part of the
phylogenetic workflow.

¹ <https://github.com/nextstrain/auspice/blob/dae78e74d5e06303d901387adf9fe7550a7d4aa9/src/components/download/helperFunctions.js#L326>

## Related issue(s)

First step for https://github.com/nextstrain/seasonal-flu/issues/178

## Checklist

- [ ] Checks pass

### DIFF
--- a/nextclade/Snakefile
+++ b/nextclade/Snakefile
@@ -415,7 +415,7 @@ rule export:
         date=datetime.datetime.utcnow().strftime("%Y-%m-%d"),
     shell:
         """
-        tsv-select -H -f {params.select_fields} {input.metadata} > {input.metadata}.tmp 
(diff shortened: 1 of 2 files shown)


huddlej (approved):

joverlee521:
I don't think there's any rush to merge this, so I'll plan to merge after I'm back from vacation to monitor this and https://github.com/nextstrain/seasonal-flu/issues/178.

PR #183, opened 2024-10-14
Upload sequences for all segments to S3

huddlej:

Adds all remaining gene to the list of segments whose sequences should be uploaded to S3.

## Checklist

- [x] Checks pass
- [x] [Test upload action from branch](https://github.com/nextstrain/seasonal-flu/actions/runs/11332980091)


huddlej:
The test upload ran without issues for all four lineages and all eight segments. The runtime was 2 hours instead of 45-50 minutes, though. We could reduce the runtime by dropping Yam from the lineages to consider. I'll push a commit next that makes this change.

joverlee521 (approved):
I was slightly worried that the additional segments might overload RethinkDB. However,  we are using the [Docker runtime](https://github.com/nextstrain/seasonal-flu/blob/208a6034aab0cdd26b4f942d28e189e3aa5a2d57/.github/workflows/upload.yaml#L47) within the upload workflow, so we are still limited to just 4 concurrent connections to Fauna.

If we are worried about the additional runtime, we could consider running the additional segments to a separate GH Action workflow that doesn't affect the automated builds.

huddlej:
> If we are worried about the additional runtime, we could consider running the additional segments to a separate GH Action workflow that doesn't affect the automated builds.

That's a good idea. Maybe we can try it this Thursday and see if it is annoyingly slow even without Yam?

This runtime highlights again how incredibly slow the RethinkDB setup is. It shouldn't take _hours_ to download this much data...

PR #192, opened 2024-10-24
Include original GISAID strain name in metadata

huddlej:

 - [x] Add the GISAID strain name to list of fields to fetch from fauna in and parse from the raw sequences FASTA in the "upload" workflow. This change will include these original names in the metadata in S3.
 - [x] Export GISAID strain names as a metadata column in Auspice JSONs
 - [x] Set GISAID strain names as default tip label

## Checklist

- [x] Checks pass

### DIFF
--- a/config/h1n1pdm/ha/auspice_config.json
+++ b/config/h1n1pdm/ha/auspice_config.json
@@ -210,7 +210,8 @@
   ],
   "display_defaults": {
     "map_triplicate": true,
-    "color_by": "clade_membership"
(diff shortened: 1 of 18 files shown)


huddlej:
Below is an example view before and after setting the default tip label to the GISAID strain.

Before:
image

After:
image

huddlej:
I'm going to merge this to include GISAID strain names in the weekly ingest, but if we decide that the default tip label change here is a bad idea, it is easy to revert for future builds.

PR #197, opened 2024-11-18
Add Nextclade extension to public builds

huddlej:

Adds Nextclade extension with "subclade" annotation to Auspice config for all public builds, allowing users to get these subclade annotations from Nextclade for any public tree.

To test this functionality, I added subclade annotations to the CI build and confirmed that the extension worked as expected by uploading [the CI Auspice JSON to the staging server](https://nextstrain.org/staging/ci/build/ha) before and after I included the extension in the JSON.

## Checklist

- [x] Checks pass
- [x] Confirm that Nextclade annotations from CI build on staging show subclade

### DIFF
--- a/config/h1n1pdm/auspice_config.json
+++ b/config/h1n1pdm/auspice_config.json
@@ -158,5 +158,16 @@
   "metadata_columns": [
     "accession_ha",
     "accession_na"
-  ]
+  ],
+  "extensions": {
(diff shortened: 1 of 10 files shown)


joverlee521 (approved):
I was wondering what happens if you include the extension in dataset without `subclade` labels and looks like you just get a blank column which seems fine.

Issue #196, opened 2024-11-12
Annotate GIHSN strains


joverlee521:
### Context

[Slack context](https://bedfordlab.slack.com/archives/C03KWDET9/p1731410466692009)

### Description

Strains from the [Global Influenza Hospital Surveillance Network (GIHSN)](https://gihsn.org/) have `GIHSN` in the strain name. We should add a rule to add a `GIHSN` column to the metadata to indicate whether the label is in the strain name. 

I'm imagining this to be a step before/after the rule to [annotate the reference strains](https://github.com/nextstrain/seasonal-flu/blob/37bcc6969b6161e2aa1e9181c07429c9cb6c79af/workflow/snakemake_rules/select_strains.smk#L94-L116).

PR #198, opened 2024-11-20
Add rule to annotate GIHSN samples

joverlee521:

Adds a new `gihsn_sample` column to the metadata to indicate whether
"GIHSN" was found in the `strain` as proxy for whether the sample came
from the Global Influenza Hospital Surveillance Network (GIHSN).
Follows the existing pattern of using `True` and `False` boolean values.

## Related issue(s)

Resolves <https://github.com/nextstrain/seasonal-flu/issues/196>

## Checklist

- [x] Checks pass

### DIFF
--- a/workflow/snakemake_rules/select_strains.smk
+++ b/workflow/snakemake_rules/select_strains.smk
@@ -70,6 +70,27 @@ rule join_metadata:
             --output {output.metadata:q} 2>&1 | tee {log}
         """
 
+# Annotate strains in the metadata that have "GIHSN" in the strain name to
+# indicate whether it was collected as part of the Global Influenza Hospital
+# Surveillance Network (GIHSN)
+rule annotate_metadata_with_gihsn:
+    input:
+        metadata="data/{lineage}/metadata_joined.tsv",
+    output:
+        metadata="data/{lineage}/metadata_with_gihsn.tsv",
+    conda: "../envs/nextstrain.yaml"
+    benchmark:
+        "benchmarks/annotate_metadata_with_gihsn_{lineage}.txt"
+    log:
+        "logs/annotate_metadata_with_gihsn_{lineage}.txt"
+    shell:
+        """
+        csvtk --tabs mutate2 \
+            --expression '${{strain}}=~"(GIHSN)" ? "True" : "False"' \
+            --name gihsn_sample \
+            {input.metadata} > {output.metadata}
+        """
+
 rule build_reference_strains_table:
     input:
         references="config/{lineage}/reference_strains.txt",
@@ -96,7 +117,7 @@ rule build_reference_strains_table:
 # later.
 rule annotate_metadata_with_reference_strains:
     input:
-        metadata="data/{lineage}/metadata_joined.tsv",
+        metadata="data/{lineage}/metadata_with_gihsn.tsv",
         references="data/{lineage}/reference_strains.tsv",
     output:
         metadata="data/{lineage}/metadata.tsv",


workflow/snakemake_rules/select_strains.smk
```diff
@@ -70,6 +70,27 @@ rule join_metadata:
...
+    conda: "../envs/nextstrain.yaml"
+    benchmark:
+        "benchmarks/annotate_metadata_with_gihsn_{lineage}.txt"
+    log:
+        "logs/annotate_metadata_with_gihsn_{lineage}.txt"
+    shell:
+        """
+        csvtk --tabs mutate2 \
+            --expression '${{strain}}=~"(GIHSN)" ? "True" : "False"' \
+            --name gihsn_sample \
```
joverlee521, line 90: I'm not tied to the `gihsn_sample` column name, so feel free to suggest any others.

huddlej (commented):
Thank you for taking this one, @joverlee521! Seeing the implementation clarifies my original concerns about adding such a specific column to the metadata when the column represents <1% of all HA sequences and the same information exists in the strain name already.

Maybe we could figure out what the specific uses for this column would be before committing to this change? I can imagine two main use cases:

1. Users want to filter for sequences from the GIHSN in their subsampling logic for their builds, so they either can include those strains or exclude them. For this use case, users could use an argument like `--query "'GIHSN' in strain"` to select for those strains.
2. Users want to filter/color sequences in a tree by the GIHSN status. For this case, users would need an additional column in the metadata like the one from this PR to enable filtering/coloring in Auspice. If users are running their own builds, they could use a custom rule based on the same code in this PR to add the column to their subsampled metadata just before augur export runs (we could help them with this). If users want to apply these filters/colorings to our public builds, then we could use the same approach of adding the custom metadata column in the workflow just before export.

If the use cases are more about selecting/filtering sequences by sampling strategy, GISAID has a "sampling strategy" column in their metadata. It isn't populated for these GIHSN samples, but we could implement a derived version of that column for which we populate empty strings with "GIHSN", for example.

@rneher Are there other reasons we might need to have GIHSN as a column in our metadata for all samples?

rneher:
I think having this as a column is probably easiest. This should compress very well and we are only talking about a hundred of thousand, rather than millions of rows.

rneher:
I think the chosen column name is fine. this should help us to include them preferentially into the build and add a coloring.

joverlee521:
Thanks both! I'll merge this so that we'll have the new column after Thursday's update.

workflow/snakemake_rules/select_strains.smk
```diff
@@ -70,6 +70,27 @@ rule join_metadata:
...
         """
 
+# Annotate strains in the metadata that have "GIHSN" in the strain name to
+# indicate whether it was collected as part of the Global Influenza Hospital
+# Surveillance Network (GIHSN)
+rule annotate_metadata_with_gihsn:
+    input:
+        metadata="data/{lineage}/metadata_joined.tsv",
+    output:
+        metadata="data/{lineage}/metadata_with_gihsn.tsv",
```
genehack, line 80: do you want to `temp()` this?

genehack (approved):

workflow/snakemake_rules/select_strains.smk
joverlee521: (hah, you're comment popped up right when I clicked merge)

Yeah, adding `temp()` is probably best here.

genehack: > (hah, you're comment popped up right when I clicked merge)

Achievement unlocked: approve PR after it is already merged 🙌

joverlee521: Updated in https://github.com/nextstrain/seasonal-flu/pull/200

PR #203, opened 2025-01-15
download_from_fauna: Add `--prioritized_seqs_file` option 

joverlee521:

Use the optional `prioritized_seqs_file` config param to define file path for fauna's new `--prioritized_seqs_file` option for downloading prioritized sequences by strain. Each file is expected to be prioritized sequences per segment, so this follows the workflows current set up to provide file paths with the `{segment}` wildcard in the config.

Includes skeleton files for all segments of h1n1pdm, h3n2, and vic that can be filled in as we define prioritized sequences.

Depends on 
- [x] merge of https://github.com/nextstrain/fauna/pull/176
- [x] [docker-base CI for release](https://github.com/nextstrain/docker-base/actions/runs/12995455081)

## Related issue(s)

Part of https://github.com/nextstrain/fauna/issues/174

## Checklist

- [ ] Checks pass


profiles/upload/h3n2/ha/prioritized_seqs_file.tsv
```diff
@@ -0,0 +1,4 @@
+strain	accession
+A/Croatia/10136RV/2023-egg	EPI3356221
```
huddlej, line 2: Looking back at the table of different Croatia isolates, I see that all of the cell-passaged isolates have the same haplotype which is why I didn't mention a specific accession to prefer for that strain. I can imagine, though, that someone uploads another cell-passaged isolate of that strain in the future with a slightly different haplotype from the passaging and we unknowingly switch to that new version in our analyses.

Could we future-proof ourselves a bit by specifying the cell-passaged accession for Croatia, too? I would pick the isolate with the fewest passages which is EPI_ISL_19085723 (HA accession: EPI3250718 and NA accession: EPI3250717).

profiles/upload.yaml
```diff
@@ -74,6 +74,7 @@ recency:
 builds:
   h1n1pdm_2y:
     lineage: h1n1pdm
+    prioritized_seqs_file: "profiles/upload/h1n1pdm/{segment}/prioritized_seqs_file.tsv"
```
huddlej, line 77: I can see why these files live close to the `upload` profile, since that is where we are using them first, but I think they should probably live in the top-level `config/` directory (e.g., `config/h3n2/ha/prioritized_seqs_file.tsv`). This placement would make the files more discoverable alongside other lineage-specific configs and encourage their reuse in other builds. Even though we plan to get rid of fauna eventually, it's still possible we could want a profile with builds that use the download from fauna rule and these prioritized sequences.

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -47,11 +47,29 @@ def _get_virus_passage_category(wildcards):
...
 
+def _get_prioritized_seqs_file(wildcards):
+    prioritized_seqs_file = None
+    for build_name, build_params in config["builds"].items():
+        if build_params["lineage"] == wildcards.lineage:
+            prioritized_seqs_file = build_params.get('prioritized_seqs_file')
+            break
+
+    if prioritized_seqs_file is not None:
+        return prioritized_seqs_file
```
huddlej, line 58: Is there a reason to not make the default value of `prioritized_seqs_file` an empty list so you could always return that variable without the condition logic here? I'm assuming you didn't want to change the type of the variable, but just checking... 😄

huddlej (commented):
Thanks, @joverlee521! I like this approach to prioritizing the sequences with a build config option. I only had two minor blocking requests about where the priority TSVs live and inclusion of a cell-passaged Croatia accession for H3N2. Otherwise, this looks good to go.

profiles/upload.yaml
joverlee521: > This placement would make the files more discoverable alongside other lineage-specific configs and encourage their reuse in other builds

Hah, I put these in the `upload` profile for the exact opposite reason. I wanted to hide them from the general configs since they aren't used in the core workflow and might be confused as unused files.

profiles/upload/h3n2/ha/prioritized_seqs_file.tsv
joverlee521: Added in a1ec64d

profiles/upload.yaml
joverlee521: Moved to the top level config in d59e5bd

workflow/snakemake_rules/download_from_fauna.smk
joverlee521: Hah, I definitely was not thinking about typing here 😄 Updated in 9a8f09c8d385fa1d6bd00c9f4d24effeb4ec28ca

joverlee521:
Running the [upload workflow](https://github.com/nextstrain/seasonal-flu/actions/runs/12996655770) to include the prioritized sequences in builds.

joverlee521:
Confirmed public builds are using the prioritized HA/NA seqs for DC/27 and Croatia/10136RV
[Tangletree with previous build](https://next.nextstrain.org/seasonal-flu/h3n2/ha/2y@2025-01-28:seasonal-flu/h3n2/ha/2y@2025-01-20?m=div&s=A/Croatia/10136RV/2023-egg,A/DistrictOfColumbia/27/2023) shows very slight difference for Croatia/10136RV.

PR #217, opened 2025-03-14
Add script and template to automate narratives

### LINKED ISSUE #163: Use a narrative template file to stub out a date-specific narrative for a given set of dated builds
Avoid the need to update all URLs and the YAML front matter in the narrative to the date of the latest builds. We could define our own simple narrative template or use a standard template language like Jinja2. We could generate the stubbed narrative as part of the build and deploy process above, such that the date of the builds is available to plug into the template, and deploy the dated draft of the narrative along with the builds.

Parts of the current report that need to be templated include:

 - title in the Markdown frontmatter (`Nextstrain seasonal influenza report for April 2024` -> `Nextstrain seasonal influenza report for {long month} {year}`)
 - date for the report (same as date of the build, `April 15, 2024` -> `{long month} {day}, {year}`) in the frontmatter
 - dataset URLs with date in the dataset and month/year filter in the URL params (`https://nextstrain.org/groups/nextflu-private/flu/seasonal/2024-04-11/h1n1pdm/2y/ha?c=gt-HA1_120,216&d=tree,map,frequencies&f_year_month=2023-10,2023-11,2023-12,2024-01,2024-02,2024-03,2024-04&gmax=1052&gmin=72&label=clade:5a.2a&p=grid`)
 - counts of recent submissions per clade (generated in separate Markdown files by the workflow)
 - tables of derived haplotype frequencies and titer references (generated in separate Markdown files by the workflow)

huddlej:

Adds a script to load a Markdown template written in [Jinja2 syntax](https://jinja.palletsprojects.com/en/stable/templates/) and write out a narrative stub with the correct dates in the narrative frontmatter and the Nextstrain URLs. The included Markdown template assumes that the narrative will be written to a `narratives/path.md` path with images in the same repo's `figures/` directory.

This PR also adds a rule to run this script from the workflow to automate narrative creation during the nextflu-private builds. We could optionally upload the narrative stub to the desired Nextstrain Group, in the future, but for now this implementation will allow us to quickly get a new report going without manually updating dates or markdown tables.

Note that Jinja2 is available in the base Nextstrain environments as a dependency for [Snakemake's own reporting functionality](https://snakemake.readthedocs.io/en/stable/snakefiles/reporting.html).

## Related issue(s)

Closes #163

## Checklist

- [x] Checks pass

### DIFF
--- a/scripts/create_narrative.py
+++ b/scripts/create_narrative.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+import argparse
+import datetime
+import sys
+
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+import pandas as pd
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument("--template", required=True, help="Template of Markdown narrative in Jinja2 format.")
+    parser.add_argument("--earliest-date", help="Earliest date in format of YYYY-MM-DD to use for year/month filters in Nextstrain views. Defaults to 6 months before the given report date.")
+    parser.add_argument("--date", help="Date in the format of YYYY-MM-DD to use for the report. Defaults to the current date.")
+    parser.add_argument("--markdown-includes", nargs="+", help="One or more key/value pairs of variable name and path to a Markdown file to be included in the template environment in the format of 'variable_name=path_to_markdown_file.md'.")
+    parser.add_argument("--output", required=True, help="Rendered Markdown narrative.")
+
+    args = parser.parse_args()
+
+    env = Environment(
+        loader=FileSystemLoader("."),
+        autoescape=select_autoescape()
+    )
+
+    template = env.get_template(args.template)
+
+    if args.date:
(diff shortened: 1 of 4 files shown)


scripts/create_narrative.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+import argparse
+import datetime
+import sys
+
+from jinja2 import Environment, FileSystemLoader, select_autoescape
+import pandas as pd
```
genehack, line 7: I see you added `jinja2`, but should probably add `pandas` to the `workflow/envs/nextstrain.yaml` file too?

genehack (approved):

scripts/create_narrative.py
huddlej: Fair point. I added jinja2 because it isn't otherwise available as a dependency of other dependencies, but pandas is a dep of augur, so I've never explicitly included it here.

huddlej: Following my own logic above, I just removed IQ-TREE and mafft dependencies from the environment, too, and bumped Augur and Nextclade versions to latest. Thanks, @genehack!

PR #218, opened 2025-03-14
Add ingest workflow

joverlee521:

Adds the basic ingest workflow for curating manually downloaded GISAID data that tries to replicate the curation done in fauna/vdb/flu_upload.py as separate curation scripts. Since we depend on the strain name to match with titer data, I've tried to keep strain name parsing as close as possible to the strain name parsing done in fauna. There are definitely still some differences in geolocation values during my own testing, but we can tweak those geolocation rules later.

Steps to test

1. Manually download the metadata and sequences from GISAID EpiFlu
2. Save the XLS metadata file as `ingest/data/2025-04-08-metadata.xls`.
3. Make sure the sequences includes all segments and the FASTA header is set to “DNA Accession no.”.
4. Save the FASTA file as `ingest/data/2025-04-08-sequences.fasta`.
5. Assuming that you downloaded the relatively new data _without_ yam, run
    ```
    nextstrain build ingest --config 'lineages=["h1n1pdm", "h3n2", "vic"]'
    ```
6. Move or copy the ingest results to the top level data directory (`cp -r ingest/results/* data/`)
7. Run a phylogenetic build. 
    I've tried the gisaid build 
    ```
    nextstrain build . --configfile profiles/gisaid/builds.yaml
    ```
    and the nextstrain-public build (which still requires credentials for titer data)
    ```
    nextstrain build --envdir ../env.d/seasonal-flu/ . auspice/seasonal-flu_h3n2_2y_na.json --configfiles profiles/nextstrain-public.yaml
    ```

## Checklist

- [ ] Checks pass


j23414:
### Ingest and subsequent phylogenetic workflows work during my testing

Works great as far as I can tell! I was able to run ingest, we chatted that I just needed to open and close the FASTA sequences. 

Linking the ingested (curated) data to the phylogenetic workflow was painless once we clarified to include at least one reference sequence when running the phylogenetic workflows. The 
subsequent Auspice JSON trees look as expected.

Looking through the commits, I see there’s a new `is_reference` column which I agree is more straight-forward, and I assume the geolocation rules will need to continuously be updated. Your rescuing location from strain name when location field is empty makes sense

No blockers from me!

joverlee521:
Thanks for testing this out @j23414! I'll merge now and continue working on the automation in https://github.com/nextstrain/seasonal-flu/pull/224.

PR #219, opened 2025-03-28
Run Nextclade in GHA for uploads

huddlej:

Adds a job to the upload GitHub Action to run Nextclade on all sequences before triggering downstream workflows. This change ensures that Nextclade annotations in S3 reflect the latest data when other workflows run. This commit removes the option to trigger the separate Nextclade GHA from the upload action. Thanks to @tsibley, this implementation reuses the Nextclade GHA, so we can still manually trigger just that action later when we need to.

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/upload.yaml
+++ b/.github/workflows/upload.yaml
@@ -58,7 +58,7 @@ jobs:
     needs: [upload]
     permissions:
       id-token: write
-    uses: nextstrain/seasonal-flu/.github/workflows/run-nextclade.yaml
+    uses: nextstrain/seasonal-flu/.github/workflows/run-nextclade.yaml@master
     secrets: inherit
 
   trigger:
--- a/.github/workflows/upload.yaml
+++ b/.github/workflows/upload.yaml
@@ -28,14 +28,6 @@ on:
         description: "Specific container image to use for nextflu-private group builds"
         required: false
         type: string
-      triggerNextclade:
-        description: "Trigger run-nextclade workflow"
-        required: true
-        type: boolean
-      nextcladeDockerImage:
-        description: "Specific container image to use for the Nextclade workflow"
-        required: false
-        type: string
 
 concurrency:
   group: ${{ github.workflow }}
@@ -62,8 +54,15 @@ jobs:
           upload_all_metadata \
           --configfile profiles/upload.yaml
 
-  trigger:
+  run-nextclade:
     needs: [upload]
+    permissions:
+      id-token: write
+    uses: nextstrain/seasonal-flu/.github/workflows/run-nextclade.yaml
+    secrets: inherit
+
+  trigger:
+    needs: [upload, run-nextclade]
     strategy:
       matrix:
         include:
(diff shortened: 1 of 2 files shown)


.github/workflows/upload.yaml
```diff
@@ -62,8 +54,15 @@ jobs:
           upload_all_metadata \
           --configfile profiles/upload.yaml
 
-  trigger:
+  run-nextclade:
     needs: [upload]
+    permissions:
+      id-token: write
+    uses: nextstrain/seasonal-flu/.github/workflows/run-nextclade.yaml
```
tsibley: I omitted it in Slack with my suggestion, but a ref is required for `uses:`, e.g.

```suggestion
    uses: nextstrain/seasonal-flu/.github/workflows/run-nextclade.yaml@master
```

But upon further consideration, since the workflows exist in the same repo as each other, it'd really be best to use

```suggestion
    uses: ./.github/workflows/run-nextclade.yaml
```

so that `run-nextclade.yaml` is used at the same revision that this calling workflow is using (see [docs](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#calling-a-reusable-workflow)).

.github/workflows/upload.yaml
```diff
@@ -76,9 +75,6 @@ jobs:
           - trigger: ${{ inputs.triggerNextfluPrivate }}
             workflow: run-nextflu-private-builds.yaml
             image: ${{ inputs.nextfluPrivateDockerImage }}
-          - trigger: ${{ inputs.triggerNextclade }}
-            workflow: run-nextclade.yaml
-            image: ${{ inputs.nextcladeDockerImage }}
```
tsibley, line 81: `inputs.nextcladeDockerImage` could still be passed thru to `run-nextclade.yaml` if you'd like.

.github/workflows/upload.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
     needs: [upload]
     permissions:
       id-token: write
-    uses: nextstrain/seasonal-flu/.github/workflows/run-nextclade.yaml
+    uses: nextstrain/seasonal-flu/.github/workflows/run-nextclade.yaml@master
```
tsibley, line 61: oh, ha.

tsibley (changes requested):
`run-nextclade.yaml` also needs a `workflow_call` triggering event added to its `on:` block, see [docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_call).

.github/workflows/upload.yaml
huddlej: > ...since the workflows exist in the same repo as each other, it'd really be best to use...

Thanks! I prefer the second option, too.

.github/workflows/upload.yaml
huddlej: I prefer to keep the interface simpler...if we need a different image here we can run the separate GHA workflow with that image.

PR #220, opened 2025-04-02
Add citations for epitope or map sites

j23414:

Add citations for the epitope of map sites in a top-level README file, listed in chronological order of publication. I put them in a table, split by type and gene, but feel free to reword or suggest a different format.

Eventually, based on [a slack thread](https://bedfordlab.slack.com/archives/C03KWDET9/p1743617647803209?thread_ts=1743617294.221779&cid=C03KWDET9), we may want to switch to adding these as DOI keys in the top-level JSON.

I wasn't confident I could safely add new keys to the JSON files without breaking something in the workflow, hence the README docs for now.

## Checklist

- [ ] Checks pass

### DIFF
--- a/config/distance_maps/README.md
+++ b/config/distance_maps/README.md
@@ -1,0 +1,30 @@
+# Distance Maps
+
+This folder contains auxillary files for distance maps. Key positions were pulled from the following publications to create the different map files.
+
+## H1N1pdm
+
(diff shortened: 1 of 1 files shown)


huddlej (approved):
Great! Thank you for making this table, @j23414!

PR #221, opened 2025-04-04
Add user-defined haplotype assignment

huddlej:
### Overview
This PR adds logic to annotate user-defined haplotypes in the phylogenetic workflow using Nextclade annotations and adds corresponding configuration files to define "emerging haplotypes" for all three subtypes. This approach allows us to define groups of substitutions that appear to be important without

1. requiring these substitutions to be grouped monophyletically (e.g., H3N2 HA strains with HA1:145N may be important regardless of their phylogenetic background)
2. releasing a new official Nextclade dataset to generate the annotations across our workflows

In this way, we can track interesting groups of viruses more flexibly by editing the haplotype definitions in this one repository.

Below is an example of the "emerging haplotypes" for H3N2 HA with J.2:145N tips hovered in the color legend to highlight their recurrent nature across the tree.

image

### Haplotype assignment script
Most of the new logic lives in a script in `scripts/assign_haplotypes.tsv` that is also a prototype interface for a future `augur haplotypes` subcommand. Additional logic lives in the core phylogenetic workflow. This script accepts a TSV input with Nextclade annotations (either the direct output from Nextclade or a merged metadata/Nextclade file) and a TSV of haplotype definitions and outputs a copy of the Nextclade TSV with the haplotype annotations in a new column. The script optionally outputs a node data JSON of the same annotations for use in phylogenetic workflows. Users can specify the id columns of the input annotations, the column to use for Nextclade “clade” definition (e.g., “clade”, “subclade”, etc.), the name of the new haplotype annotation column, the default haplotype value to assign when there isn’t a match to any definition, and a boolean flag to set the default haplotype to the current record’s clade value. We run the script like this:

```
python scripts/assign_haplotypes.py \
    --substitutions nextclade.tsv \
    --haplotypes config/h3n2/ha/emerging_haplotypes.tsv \
    --clade-column subclade \
    --haplotype-column-name emerging_haplotype \
    --output-table emerging_haplotypes.tsv
```

For phylogenetic workflows, we can also output a node data JSON to include in `augur export` inputs like so:

```
python scripts/assign_haplotypes.py \
    --substitutions nextclade.tsv \
    --haplotypes config/h3n2/ha/emerging_haplotypes.tsv \
    --clade-column subclade \
    --haplotype-column-name emerging_haplotype \
    --output-table emerging_haplotypes.tsv \
    --output-node-data emerging_haplotypes.json
```

### Haplotype definitions
Users define haplotypes in a TSV file using the a format similar to the one used to define clades for `augur clades`. The following is an example of the H3N2 HA haplotypes. This syntax allows users to define haplotypes based on existing clades with or without additional nucleotide or amino acid substitutions that have emerged on that background. The haplotype assignment script uses the fact that Nextclade produces a list of these substitutions relative to a “founder” to easily find matches based on the existing clade and list of additional substitutions.

```
haplotype	gene	site	alt

# Define haplotypes based on existing clades.
J.2	clade	J.2
J.2.1	clade	J.2.1
J.2.2	clade	J.2.2

# Define haplotypes based on existing clades with one additional substitution.
J.2:8D	clade	J.2
J.2:8D	HA1	8	D

J.2:145N	clade	J.2
J.2:145N	HA1	145	N

J.2:135K	clade	J.2
J.2:135K	HA1	135	K

J.2:223I	clade	J.2
J.2:223I	HA1	223	I

# Define haplotypes based on existing clades with multiple additional substitution.
# If sequences have multiple matching haplotypes, they receive the last haplotype in
# this definition file.
J.2:124R-145N	clade	J.2
J.2:124R-145N	HA1	124	R
J.2:124R-145N	HA1	145	N

J.2:135A-145N	clade	J.2
J.2:135A-145N	HA1	135	A
J.2:135A-145N	HA1	145	N

J.2:135K-189R	clade	J.2
J.2:135K-189R	HA1	135	K
J.2:135K-189R	HA1	189	R

J.2:145N-158K	clade	J.2
J.2:145N-158K	HA1	145	N
J.2:145N-158K	HA1	158	K

J.2:158K-189R	clade	J.2
J.2:158K-189R	HA1	158	K
J.2:158K-189R	HA1	189	R
```

Users can also define haplotypes using only substitutions without any reference to existing clades. For example, the following file redefines subclade J.2 based on its defining substitutions.

```
haplotype	gene	site	alt
J.2	HA1	122	D
J.2	HA1	276	E
```

My plan is for these definitions to live in the seasonal-flu repository where they can be fetched by other workflows that also need to annotate haplotypes (e.g., frequency estimation workflows). Those workflows already download Auspice config JSONs to get clade colors, so they just need additional logic to download the haplotype definitions and use them to generate annotations.

For records that don't match any haplotype definitions, the `assign_haplotype.py` script provides two different ways to set a "missing" value. By default, the script will assign records a value based on the script's `--default-haplotype` value ("unassigned" by default). In the context of frequency estimation workflows, we can run the script with `--default-haplotype other` to bin all records without a defined haplotype into the "other" category that we use for SC2 frequency estimation.

The second option is to run the script with the `--use-clade-as-default-haplotype` flag which will set the haplotype of each record to its Nextclade clade assignment when there is no matching haplotype definition. In the context of our monthly reporting about new sequences, this flag allows us to easily count new sequences by haplotype including sequences that don't map to any of the haplotypes we are actively tracking.

### Limitations
Haplotype assignments require a Nextclade TSV with subclade, nucleotide substitutions, and amino acid substitutions, but the core phylogenetic workflow does not produce these annotations when it runs Nextclade for alignment. To get those annotations, we would need to run Nextclade with an official dataset or Auspice tree JSON input. In the long term, we probably want to move toward that approach, but that would require a bigger change to the workflow. As a short-term solution, this PR downloads existing Nextclade annotations from S3, filters those records to match the strains selected for each build, and assigns haplotypes based on the subset of annotations. This approach only works for the Nextstrain team. External users will have to wait until we've implemented a solution based on Nextclade annotations produced by the workflow itself.

The current haplotype assignment script and workflow logic lack the the ability to annotate haplotypes on internal nodes. This means we can’t get branch labels that we can access via Auspice URLs in narratives. To get these branch labels, we could generate our own Nextclade annotations in the workflow for all tips and internal node sequences by running Nextclade a second time with the sequence output from `augur ancestral` and an official Nextclade dataset that provides clade and substitution columns. With those annotations, we could modify the current haplotype assignment script to generate a `branch_labels` entry in its node data JSON for internal nodes that have the first occurrence of each haplotype.

Finally, we already have two different "derived haplotype" scripts and rules in this workflow which makes this new functionality more confusing in the context of the whole code base. We should revisit the value of all these various approaches and consolidate to fewer options. We could rename these annotations to something like “manual haplotypes” and “automated haplotypes” to disambiguate their origins and purposes.

### Try it out

To run the workflow from your laptop with simple example data, run the example build with the following command.

```
nextstrain build .  --configfile profiles/example/builds.yaml
```

When the build has finished running, view the output Auspice trees via:

```
nextstrain view auspice/
```

Haplotypes are defined in the file `example_data/haplotypes.tsv`.

## Checklist
- [x] Checks pass


scripts/assign_haplotypes.py
```diff
@@ -0,0 +1,227 @@
...
+            clade_match = True
+            nucleotide_column = "substitutions"
+            aa_column = "aaSubstitutions"
+
+        if (
+            clade_match and
+            nucleotide_substitutions_match(record.get(nucleotide_column, ""), definition.get("nuc", [])) and
+            aa_substitutions_match(record.get(aa_column, ""), definition.get("aa", []))
+        ):
+            assigned_name = name
```
jameshadfield, line 118: This will allow overwriting of the assigned haplotype if multiple matches are in the definitions TSV. This is what we want really - e.g a strain which we want to assign `J.2:8D` will first be assigned `J.2` then overwritten. But it's behaviour [that's tripped us up](https://github.com/nextstrain/augur/issues/1711) in `augur clades`. I don't have a good solution here, other than an understanding that the order of ones definitions TSV is very important.

scripts/assign_haplotypes.py
```diff
@@ -0,0 +1,227 @@
...
+        else:
+            # Try to assign this haplotype based on all substitutions.
+            clade_match = True
+            nucleotide_column = "substitutions"
+            aa_column = "aaSubstitutions"
+
+        if (
+            clade_match and
+            nucleotide_substitutions_match(record.get(nucleotide_column, ""), definition.get("nuc", [])) and
+            aa_substitutions_match(record.get(aa_column, ""), definition.get("aa", []))
```
jameshadfield, line 116: This code stood out for me - under what circumstances are these columns missing from individual records? I would have thought the columns will always be here or the program should error out...

scripts/assign_haplotypes.py
```diff
@@ -0,0 +1,227 @@
...
+        clade_column=args.clade_column,
+        default_haplotype=args.default_haplotype,
+        use_clade_as_default_haplotype=args.use_clade_as_default_haplotype,
+    )
+
+    # Assign haplotypes to each row.
+    substitutions[args.haplotype_column_name] = substitutions.apply(
+        assign_haplotype_per_record,
+        axis=1,
+    )
```
jameshadfield, line 213: [question] Did you gravitate to pandas here (because of speed, ergonomics etc) or just continue to use it because `read_metadata` returns a dataframe?

jameshadfield (commented):
The following commentary is about the general `augur haplotypes` usage, not seasonal-flu specific. Overall I think will be very useful, and nicely addresses  [an old comment of mine from 2021](https://github.com/nextstrain/augur/issues/735):

> The behaviour of augur clades means that if there are multiple nodes containing clade-defining mutations (i.e. the clade is polyphyletic), then we only annotate clades on the biggest monophyly. We should warn when situations like this arise, or allow this to be relaxed. I expect it'll become common to want to define "clades" via a small set of constellation nCoV mutations, and expect polyphyletic colourings in Auspice.

### Clade is an overloaded word!

The augur clades TSV is entirely self-referential, so that all clades must be defined within the TSV itself, and then other clade definitions can build off previously defined clades. (It's actual implementation isn't great as it doesn't enforce the implicit heirarchical relationship, but we can/will change that.) With haplotypes I found this more hazy... as an example:

```
# Define haplotypes based on existing clades.
J.2	clade	J.2
J.2 HA1	42	A          # my addition

# Define haplotypes based on existing clades with one additional substitution.
J.2:8D	clade	J.2
J.2:8D	HA1	8	D
```


My initial thought - is `J.2:8D` leveraging the clade J.2 as defined in the metadata TSV, or as defined earlier in this definitions TSV? I had to read the code to understand it's the former.

Secondly, when we see `XXX	clade	J.2` it seems like we should be matching the metadata TSV on `clade=J.2` , but in reality we are actually looking for `subclade=J.2` due to `--clade-column subclade`. Why not avoid this indirection by specifying `XXX	subclade	J.2` in the TSV? 

Thirdly, because you leverage Nextclade's mutations vs. founder you are enforcing hierarchy which `augur clades` currently does not -- nice! -- but you will miss matches for a strain which nextclade assigned a subclade other than (e.g.) `J.2` , e.g. perhaps a child subclade of `J.2`.


### Internal branches

> To get these branch labels, we could generate our own Nextclade annotations in the workflow for all tips and internal node sequences by running Nextclade a second time with the sequence output from augur ancestral and an official Nextclade dataset that provides clade and substitution columns. With those annotations, we could modify the current haplotype assignment script to generate a branch_labels entry in its node data JSON for internal nodes that have the first occurrence of each haplotype.

Intuitively this seems like we should traverse mutations on a tree here rather than calling nextclade call on reconstructed nodes. I realise there's some subtlety around how a "clade" would be defined in the tree. There's also some subtlety around branch labels for non-monophyletic assignments. Either way, that's all beyond this PR.

### Should we add this to `augur clades`?

This doesn't block this PR in the slightest (of course!) but I can see a path where this is part of `augur clades` rather than a separate command, especially if we start labelling internal branches. We can discuss again after it's seen some use in seasonal-flu...

scripts/assign_haplotypes.py
huddlej: Both because I had this script in mind as a new Augur command where `read_metadata` is the standard interface and for the ergonomics of the `apply` method.

huddlej:
Thank you for the review, @jameshadfield! I've responded to main points below. Maybe we could chat synchronously about this today or tomorrow?

> The augur clades TSV is entirely self-referential, so that all clades must be defined within the TSV itself, and then other clade definitions can build off previously defined clades. (It's actual implementation isn't great as it doesn't enforce the implicit heirarchical relationship, but we can/will change that.) With haplotypes I found this more hazy...
>
> My initial thought - is J.2:8D leveraging the clade J.2 as defined in the metadata TSV, or as defined earlier in this definitions TSV? I had to read the code to understand it's the former.

I see what you mean! I wanted to reuse the existing clades TSV approach here as much as possible to avoid defining a new format for users (including us) to learn, but I see how familiarity with the original `augur clades` behavior will confuse people in this context. One simple solution would be to remove the ability to reference existing clades to define haplotypes, forcing users to define haplotypes with specific alleles for nucleotides and amino acids. This solution would also address the issue you mention below about missing matches within subclades of clades (e.g., missing a match in J.2.2 because we’ve specified J.2 as the clade). On the other hand, this solution throws out a really nice feature of Nextclade which is the ability to refer to substitutions relative to a founder clade.

What if we used a similar syntax to define haplotypes but with a clearer set of headings? Like you mention in another comment, we could make the value of the `gene` column refer to the Nextclade column we want (like `subclade`) instead of hardcoding `clade`. We could also change the header line to look like this:

```
haplotype	gene	site	alt
J.2	subclade	J.2
J.2.1	subclade	J.2.1
J.2.2	subclade	J.2.2

J.2:8D	subclade	J.2
J.2:8D	HA1	8	D
```

The different behavior of these definitions is one reason why I’d think to keep this implementation separate from `augur clades`. Do you think that difference in definition syntax plus the use of a different command like `augur haplotypes` would provide enough context to minimize potential confusion for users who think haplotypes get defined hierarchically?

> because you leverage Nextclade's mutations vs. founder you are enforcing hierarchy which augur clades currently does not -- nice! -- but you will miss matches for a strain which nextclade assigned a subclade other than (e.g.) J.2 , e.g. perhaps a child subclade of J.2.

This is a big gotcha for sure. I’ve already made this mistake in my own use of this script where I’m considering J.2:145N but missing any 145N haplotypes in J.2.1, J.2.2, etc. This is a context-specific problem for this one pathogen, but it could be a general issue for people. The simplest solution is to fallback to a clade-less haplotype definition (e.g., based on 145N alone). A more complicated solution would be to allow multiple subclades to be listed for a single haplotype like either of these:

```
# Allow a list of subclades per haplotype like this:
J.2:145N	subclade	J.2,J.2.1,J.2.2
J.2:145N	HA1	145	N

# Or allow overloading the subclade entries like this:
J.2:145N	subclade	J.2
J.2:145N	subclade	J.2.1
J.2:145N	subclade	J.2.2
J.2:145N	HA1	145	N
```

In other contexts like SARS-CoV-2 pango lineages, I’ve seen this problem handled with globbing like `J.2*`, but aliasing of lineages here will cause problems (like if we renamed `J.2.2` to `K`).

> Internal branches
> …that's all beyond this PR.

Agreed :D My main thought here is that our existing “branches” implementation in node data JSONs doesn’t make sense with non-monophyletic annotations anyway.

> I can see a path where this is part of augur clades rather than a separate command, especially if we start labelling internal branches

I think I see what you mean, but I’m worried about confusing users by putting different-but-similar functionality into the same command. It would be especially confusing to allow non-clade assignment in a command called `clades`. I also agree that maybe something like this implementation would eventually workaround current issues with `augur clades`, though.

From [your code comment](https://github.com/nextstrain/seasonal-flu/pull/221#discussion_r2053138562):
> This will allow overwriting of the assigned haplotype if multiple matches are in the definitions TSV. This is what we want really - e.g a strain which we want to assign J.2:8D will first be assigned J.2 then overwritten. But it's behaviour [that's tripped us up](https://github.com/nextstrain/augur/issues/1711) in augur clades. I don't have a good solution here, other than an understanding that the order of ones definitions TSV is very important.

You’re totally right. The order matters a lot and there is nothing in the definition or the script’s UI to communicate that to users. We could solve this by introducing a `priority` column to the haplotype (and clade) definitions where users could specify a numeric priority that the script will internally sort by before assigning clades. This would ensure that we always assign the highest priority haplotype whenever there are multiple options.

I tried implementing all of the ideas above in the H3N2 haplotype definitions to get this:

```
haplotype	gene	site	alt	priority
# Absence of a priority equals priority of 0.
J.2	subclade	J.2
J.2.1	subclade	J.2.1
J.2.2	subclade	J.2.2

J.2:8D	subclade	J.2
J.2:8D	HA1	8	D

J.2:145N				1
J.2:145N	subclade	J.2,J.2.1,J.2.2
J.2:145N	HA1	145	N

J.2:135K				1
J.2:135K	subclade	J.2
J.2:135K	HA1	135	K

J.2:223I	subclade	J.2
J.2:223I	HA1	223	I

J.2:124R-145N				2
J.2:124R-145N	subclade	J.2
J.2:124R-145N	HA1	124	R
J.2:124R-145N	HA1	145	N

J.2:135A-145N				2
J.2:135A-145N	subclade	J.2
J.2:135A-145N	HA1	135	A
J.2:135A-145N	HA1	145	N

J.2:135K-189R				2
J.2:135K-189R	subclade	J.2
J.2:135K-189R	HA1	135	K
J.2:135K-189R	HA1	189	R

J.2:145N-158K				2
J.2:145N-158K	subclade	J.2
J.2:145N-158K	HA1	145	N
J.2:145N-158K	HA1	158	K

J.2:158K-189R				3
J.2:158K-189R	subclade	J.2
J.2:158K-189R	HA1	158	K
J.2:158K-189R	HA1	189	R
```

In practice, I find the `priority` column annoying to define because of its several empty columns. It’s also not clear what value I should use in `priority` column for transitive relationships like haplotypes with 145N, 145N-158K, and 158K-189R. There are actual sequences with all three substitutions 145N, 158K, and 189R. If I’ve only defined the three haplotypes list above, there isn’t an obvious high priority haplotype to assign to sequences with all three substitutions. In practice, I’d need to define another haplotype that includes all three substitutions to avoid incorrectly classifying those sequences.

I also don’t like that if we allow the `priority` column to be empty and set default priorities to 0, then when we sort the definitions in the script by priority, we will lose the original ordering in the file that users might expect to encode priority. For instance, J.2:8D doesn’t have a priority over J.2, so the script could assign sequences with the HA1:8D substitution to J.2.

Based on this prototype, I’d lean toward the priority being set by order of the definitions or by requiring a priority for every entry that could potentially conflict.

jameshadfield:
> In practice, I find the priority column annoying ... Based on this prototype, I’d lean toward the priority being set by order of the definitions

Yes, I also prefer the TSV ordering. One nice addition would be logging the overwrites, either in a "log" section within the JSON itself or on STDOUT, such as:

```
The following strains matched multiple haplotype definitions, with the last match chosen:
Strain X: J.2 → J.2:135K
...
```


> What if we used a similar syntax to define haplotypes but with a clearer set of headings? 

I like this. You could go further and consider changing `gene` to `gene|founder` in the header. 

> Or allow overloading the subclade entries like this

Yes - not sure which format I prefer, but this seems a good feature to allow. Comma separated probably best if we know (sub)clades can't use them!

PR #225, opened 2025-04-24
profiles/nexstrain-public: Remove filter `--exclude-where passage=egg`

joverlee521:

Noticed by @j23414 during testing with the new ingest results that there were warnings about the `passage` column not existing in the metadata.

Turns out this is a warning we also see in the prod builds because we do no have a `passage` column in our current metadata.

<https://github.com/nextstrain/seasonal-flu/actions/runs/14648450833/job/41108380205#step:5:2695>

## Checklist

- [ ] Checks pass


joverlee521:
Merging without review since these filters were not doing anything in the build anyways.

PR #233, opened 2025-05-28
Ingest updates

joverlee521:

Updates to the ingest workflow that I made as I compared the trial outputs with the fauna metadata.
There are still geolocation updates that need to be made that I will address separately. 

One missing chunk of data are B types that are missing "Lineage" from GISAID, so they cannot be classified as vic or yam. I'm not sure how they were originally labelled as vic or yam in fauna. This might be a future improvement where we run Nextclade in the ingest workflow to classify them rather than rely on the GISAID labels. 

## Checklist

- [x] Checks pass
- [x] [Trial run](https://github.com/nextstrain/seasonal-flu/actions/runs/15329196985)

### DIFF
--- a/ingest/defaults/config.yaml
+++ b/ingest/defaults/config.yaml
@@ -57,6 +57,7 @@ curate:
     Host: host
   gisaid_subtype_field: "Subtype"
   gisaid_lineage_field: "Lineage"
+  gisaid_note_field: "Note"
   new_type_field: "vtype"
   new_subtype_field: "subtype"
   new_lineage_field: "lineage"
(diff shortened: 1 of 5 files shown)


joverlee521:
Merging since the trial run completed successfully and the output strain names are fixed.

jameshadfield:
> One missing chunk of data are B types that are missing "Lineage" from GISAID, so they cannot be classified as vic or yam. I'm not sure how they were originally labelled as vic or yam in fauna. This might be a future improvement where we run Nextclade in the ingest workflow to classify them rather than rely on the GISAID labels.

Can we use manual override lists (derived from fauna) to preserve historical stuff like this?

joverlee521:
> > One missing chunk of data are B types that are missing "Lineage" from GISAID, so they cannot be classified as vic or yam. I'm not sure how they were originally labelled as vic or yam in fauna. This might be a future improvement where we run Nextclade in the ingest workflow to classify them rather than rely on the GISAID labels.
> 
> Can we use manual override lists (derived from fauna) to preserve historical stuff like this?

Good question! We currently cannot. I'll think on how this should work since we filter by lineage pretty early on in the curation pipeline.

PR #237, opened 2025-06-19
Add overview info to ingest READMEs

jameshadfield:
These notes helped me when reading through this work for the first time

It'd also be a good idea for us to document the process around re-processing everything (delete the cache & move processed to unprocessed, I think?)

### DIFF
--- a/ingest/README.md
+++ b/ingest/README.md
@@ -1,11 +1,15 @@
 # Ingest
 
-This workflow ingests manually downloaded data from GISAID and outputs curated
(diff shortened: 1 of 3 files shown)


joverlee521 (approved):
Thanks for writing this up @jameshadfield! 

> It'd also be a good idea for us to document the process around re-processing everything (delete the cache & move processed to unprocessed, I think?)

Definitely good to think about this now. In addition to documenting the steps, I think I'd want to add the steps to the `nextstrain-automation` workflow and have a config param to toggle on re-processing all data.

PR #240, opened 2025-07-29
Enable filtering samples by nextclade QC scores

j23414:

This PR has the long term goal to enable filtering out low quality samples before they reach the tree building step. Opening this as a draft for discussion and collaborative implementation.

**Potential TODOs (feel free to edit)**

* [x] Add rules and functions to merge Nextclade columns with the metadata used for a particular build.
* [x] Add the Nextclade QC query filter (in configs?)
* [ ] (Optional) Instead of merging all nextclade columns, select the columns necessary for filter or annotation -> maybe in a later PR

## Related issue(s)

* https://github.com/nextstrain/seasonal-flu/issues/144

## Checklist

- [x] Checks pass
- [x] Update changelog


j23414:
Pair-programming with @huddlej — thanks! Feel free to push additional changes or suggestions.

I ran the following quick checks and the results look okay to me:

```bash
# profiles/nextflu-private-forecasts.yaml
envdir ~/nextstrain/env.d/seasonal-flu/ \
  nextstrain build . \
  builds/forecasts_h3n2_2y_1/metadata.tsv \
  --configfile profiles/nextflu-private-forecasts.yaml

vd builds/forecasts_h3n2_2y_1/metadata.tsv

# profiles/nextflu-private.yaml
envdir ~/nextstrain/env.d/seasonal-flu/ \
  nextstrain build . \
  builds/h1n1pdm_2y_titers/metadata.tsv \
  --configfile profiles/nextflu-private.yaml

vd builds/h1n1pdm_2y_titers/metadata.tsv

# profiles/nextstrain-public.yaml
envdir ~/nextstrain/env.d/seasonal-flu/ \
  nextstrain build . \
  builds/seasonal-flu_vic_6m/metadata.tsv \
  --configfile profiles/nextstrain-public.yaml

vd builds/seasonal-flu_vic_6m/metadata.tsv 

# profiles/private.nextflu.org.yaml
envdir ~/nextstrain/env.d/seasonal-flu/ \
  nextstrain build . \
  builds/flu_niid_h1n1pdm_2y_cell_hi/metadata.tsv \
  --configfile profiles/private.nextflu.org.yaml

vd builds/flu_niid_h1n1pdm_2y_cell_hi/metadata.tsv
```

huddlej:
Thank you for getting this setup, @j23414! One issue with our approach here that I completely overlooked is that users outside of our group will not have a Nextclade TSV at the start of the workflow. We only expect them to prepare a metadata TSV and sequences FASTA, as described in the GISAID quickstart guide. With the changes in this PR, that GISAID quickstart guide fails at the `nextstrain build` step because the workflow can’t find a `data/h3n2/ha/nextclade.tsv.xz` file.

I pushed a couple new commits that make the workflow prefer our S3 Nextclade TSV when the `download_from_s3.smk` is being used, but otherwise the workflow will download the Nextclade dataset per subtype and segment and run Nextclade on the user's input sequences. This logic lives in the `select_strains.smk`, so the workflow can make the `data/{lineage}/{segment}/nextclade.tsv` file when it needs to.

One nice side effect of this change is that we no longer need to maintain a separate Nextclade TSV for the CI builds or define rules to select a subset of Nextclade rows to annotate emerging haplotypes; we can just use the merged metadata with Nextclade TSV for everything.

I've tested the new changes with our S3 inputs and with/without titers and I also tested the GISAID quickstart.

The only other change I made to your original commits was to remove the QC coloring from the NA Auspice configs to avoid confusion for users who might think that the coloring represents the quality of the NA sequences when it is actually only the quality of HA sequences right now.

I will push an update to the change log and then I think this is ok to merge, unless you have any other changes you'd like to make before then?

huddlej (approved):

Issue #254, opened 2025-09-24
H3N2 sequences classified as J.2.4 in Nextclade appear only as J.2 in my Nextstrain tree


RasmusKoRiis:
#Description / Problem Statement
When I analyze certain H3N2 sequences with Nextclade, they are classified into subclade J.2.4.
However, when I build a tree with Nextstrain, those same sequences are shown only as J.2 (the parent clade).

This looks like a mismatch between the clade definitions used in Nextclade vs those used in the Nextstrain seasonal influenza build.

#Steps to Reproduce / Workflow

I have used the both nextclade in CLI and online version to get J.2.4 classification of my H3N2 seqeunces. 

I then run the same H3N2 seqeunces with references using the GISAID-workflow in your nextstrain (seasonl flu) pipeline and get the seqeunces classified as J.2.

#Environment / Versions

Nextstrain CLI: 10.2.1.post1

OS: Linux (Ubuntu-based)

Expected Behavior
I expected the Nextstrain tree build to classify these sequences as J.2.4, consistent with Nextclade.

Actual Behavior
In the Nextstrain tree, they appear only as J.2.

huddlej:
Thank you for reporting this issue, @RasmusKoRiis! The seasonal-flu workflow downloads both clade and subclade annotations from the influenza-clade-nomenclature repository for each subtype and gene segment combination (for example: [this is the H3N2 HA repo](https://github.com/influenza-clade-nomenclature/seasonal_A-H3N2_HA)) and then runs `augur clades` with those downloaded files. This implementation means that the downloaded copies of the clade definitions can get stale without Snakemake realizing it. There are other possible reasons we could get mismatched results between the seasonal-flu workflow and Nextclade, but this seems like the simplest explanation to check first.

Can you share the contents of the last 5 lines of your local copy of `config/h3n2/ha/subclades.tsv`? The lines should look like this:

```
J.3	HA2	176	I
J.4	clade	J
J.4	HA1	173	R
J.4	HA1	276	E
J.4	SigPep	3	T
```

If you don't see J.4 at the end of the file, you can force the workflow to redownload the clade definitions by adding the `--forceall` Snakemake flag to the end of the command from the quickstart guide like so:

```
nextstrain build . --configfile profiles/gisaid/builds.yaml --forceall
```

huddlej:
[...]

Issue #131, opened 2023-11-28
Use Nextclade to assign clade labels in main phylogenetic workflow


huddlej:
## Context

We currently assign clade labels to trees in our main phylogenetic workflow using the `augur clades` command and the influenza clade nomenclature TSVs. However, clade assignments vary for some samples between these public/private trees and the Nextclade trees. Clade assignments can vary between different runs of public/private tress from the same time period due to different sample compositions of the trees produced by our random subsampling logic. These mismatches can cause confusion among users who look at both Nextclade outputs for their own data and the public/private Nextstrain trees.

## Description

Since Nextclade provides a standard clade label interface already, we should use Nextclade to annotate clades in our main phylogenetic workflow instead of `augur clades`. This change will ensure that the samples are assigned to the same clade regardless of the sample composition of a given public/private tree.

## Possible solutions

In the short term, we could replace our `nextalign` alignment with `nextclade` using the corresponding reference's dataset for each subtype. We would need to replace the current `augur clades` command with a version like @corneliusroemer proposed in https://github.com/nextstrain/augur/pull/1329 that allows us to assign clades to internal nodes and branches for complete backward compatibility of clade display in Auspice.

In the long (medium?) term, we could run Nextclade during our "data upload to S3" workflow, upload the alignments and Nextclade annotations joined with metadata, and then start our workflows with those files. This approach would allow us to skip the alignment and clades steps of the current workflow and it would provide useful Nextclade data on S3 that we need for other analyses like flu frequencies, etc.

huddlej:
We recently [added logic to the phylogenetic workflow to support running Nextclade on all sequences](https://github.com/nextstrain/seasonal-flu/pull/240#issuecomment-3140940748). This change allowed us to filter sequence by their QC annotation, but it also allows us to pull in Nextclade subclade annotations in the future. Before we can use those subclade annotations, we still need to decide how to best annotate subclades for internal nodes so we get the expected branch labels and colorings for those nodes.

jameshadfield:
> we still need to decide how to best annotate subclades for internal nodes so we get the expected branch labels and colorings for those nodes

[...]

PR #258, opened 2025-10-02
Export Nextclade subclade columns for all segments to build metadata

huddlej:

Annotates metadata with the subclade column from each segment's Nextclade annotations (e.g., "subclade_nextclade_ha") and exposes these new columns through the Auspice config JSONs for the HA/NA public
builds. Depending on the composition of the input sequences, augur clades' subclade annotations can be incorrect. In contrast, Nextclade's clade annotations do not depend on the collective composition of the input sequences. This implementation does not fully resolve the issue with augur clades' subclade annotations, since I have not annotated clades for internal nodes or handled branch labels. However, this approach should allow us to doublecheck augur clades annotations.

As part of the new Auspice config colorings, I chose to include both HA and NA clade colorings in both HA and NA Auspice config JSONs. This inclusion allows users to visualize potential reassortment events from a single HA or NA tree by either coloring the tree by the other segment's clade or by scatterplotting the HA and NA clades on the x and y axes.

Below is an example scatterplot for a subset of recent H1N1pdm samples showing HA and NA clades from Nextclade on x and y axis.

image

This view shows:

 1. reassortment within HA clade D.3.1 (to NA clades D.1 and D.2)
 2. misannotation of a HA D.3.1 sample as D.4 by `augur clades` (gray dot in the bottom left)
 3. misannotation of a HA D.3.1.1 sample as D.3.1 by Nextclade (orange dot in the D.3.1 column on the left)

This approach could be a first step toward the complete replacement of `augur clades` with Nextclade as described in #131. However, Nextclade's own annotations can have their own issues (as shown above) and samples from monophyletic clades can appear in polyphyletic groups in the tree complicating assignment of unique branch labels.

## Related issue(s)

Related to #254
Related to #131

## Checklist

- [x] Checks pass
- [x] Test with GISAID quickstart with HA only
- [x] Test with quickstart builds for all three lineages and HA/NA
- [x] Update changelog

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,10 @@ Changes for this project _do not_ currently follow the [Semantic Versioning rule
 Instead, changes appear below grouped by the date they were added to the workflow.
 The "__NEXT__" heading below describes changes in the unreleased development source code and as such may not be routinely kept up to date.
 
+# 9 March 2026
+
+- Export Nextclade subclade columns for each segment to the build metadata (e.g., `subclade_nextclade_ha`, `subclade_nextclade_na`, etc.), allowing any segment's tree to color tips by any other segment's clade annotation. See [#258](https://github.com/nextstrain/seasonal-flu/pull/258) for details.
+
 # 18 December 2025
 
 - Updated quickstart guide for working with GISAID data. See [#282](https://github.com/nextstrain/seasonal-flu/pull/282) for details.
--- a/config/h1n1pdm/ha/auspice_config.json
+++ b/config/h1n1pdm/ha/auspice_config.json
@@ -80,6 +80,16 @@
       ]
     },
     {
+      "key": "subclade_nextclade_ha",
+      "title": "HA subclade (Nextclade)",
+      "type": "categorical"
+    },
+    {
+      "key": "subclade_nextclade_na",
+      "title": "NA subclade (Nextclade)",
+      "type": "categorical"
+    },
+    {
       "key": "emerging_haplotype",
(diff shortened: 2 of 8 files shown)


jameshadfield (approved):
A _lot_ of TSV merging here and renaming of "clade" to "subclade" etc! One day we can do all this in ingest.

For what it's worth, I didn't understand

> Depending on the composition of the input sequences, augur
clades' subclade annotations can be incorrect. In contrast, Nextclade's
clade annotations do not depend on the collective composition of the
input sequences.

I presume here you mean that different runs change the ordering of mutations (as you traverse), or situations with multiple "clades" with (homoplasic) mutations where the largest clade varies across runs?

huddlej:
> A _lot_ of TSV merging here and renaming of "clade" to "subclade" etc!

Fair point. If all input TSVs had the same strain id, it might be clearer to do a single merge of all the TSVs with `merge_nextclade_with_metadata`. I could still do that single merge, but I'd need to make `--metadata-id-columns` arguments for each segment's Nextclade clades TSV. 🤔

Also, we should be able to drop "subclade" in favor of "clade" any time we want.

> I presume here you mean that different runs change the ordering of mutations (as you traverse), or situations with multiple "clades" with (homoplasic) mutations where the largest clade varies across runs?

I meant more that different users will run the workflow with more or less historical contextual sequences. With fewer historical contextual sequences, the ancestral nodes won't always have the mutations associated with ancestral clades and so any clade that depends on those mutations won't get assigned properly. #254 is a good example of that issue.

PR #267, opened 2025-11-26
[ingest] general curation fixes

jameshadfield:
This started out as an effort to move away from using strain names as the canonical ID for curation steps but turned into more of a grab-bag of fixes and changes. Please see the commit messages for full details. There's a few big-picture directions this PR (and subsequent PRs) aim for.

Each layer in the curation chain is intended to be self-contained where possible. For instance, any and all changes to lineages should happen in the `./scripts/standardize-lineage` layer, including hardcoded annotations. This means we have more annotations TSVs, but errors/warnings printed by layers can be fixed at the source rather than later on. This allows us to address all the warnings in the curate logs and the majority of code-changes in this commit do just this, the end result being an empty curate log; this will eventually allow us to diff curated files to understand any curation changes. 

While a motivating direction was to replace strain names with EPI ISLs, I ended up only doing this for the curate layers above ./scripts/standardize-strain-names. Specifically: lineage annotations & location annotations. Because strain names are so widely used in curation (prioritised strains, reference strains), titers, and phylo workflows (include / exclude files) we should invest in tooling which identifies when strain names _change_ so that we can keep all these files in sync.

I made a start on curation docs, but they're very much a work in progress.

The next PR will build on this to address avian-flu focused curation changes.

### DIFF
--- a/ingest/curation-docs.md
+++ b/ingest/curation-docs.md
@@ -1,0 +1,40 @@
+# Curation docs
+
+> WORK IN PROGRESS!
+
+_This guides you through how to change the rules & overrides used by the curation pipeline in order to fix metadata for records._
+
+
+### High level sketch of the curation experience
+
+* How do you know what needs fixing? Just read `logs/curate.txt` (which parallels the fauna path) or ???
+
+I think the experience goes something like:
+
+1. Download newest cache
+2. Run with the old code to get the (full) curated NDJSON & store it somewhere
+3. Make code/TSV changes, following instructions here
+4. Rerun the curate pipeline & diff against (2). We want a user-friendly diffing approach which identifies what's changed without being too verbose
+
+
+## Geography
+
+Geography fields "region", "country", "division" and "location" are set via a multi-layered approach. Here's a high-level overview, as this will help identify where a fix can/should be made:
+
+1. The `parse-gisaid-location` script parses the (capital-L) "Location" field from the GISAID record, which is expected to have slash-separated geographic information.
+  1a. `defaults/locations.tsv` sets custom fields for epi-isls in the TSV. If the record has an annotation then no more modifications are done by the script (i.e. 1b, 1d and 1d are skipped).
+  1b. `defaults/gisaid_location_rules.tsv` remaps the slash-separated Location fields. This is mostly used to correct location values with more than 4 fields so that the script can parse them.
+  1c. Sets the slash-separated Location value to "region", "country", "division" and "location"
+  1d. If four fields aren't populated then we attempt to extract a further piece of geographic information from the strain name.
+2. `augur curate titlecase` changes casing
+3. `augur curate apply-geolocation-rules`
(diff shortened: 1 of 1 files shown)


ingest/scripts/standardize-lineage
```diff
@@ -5,11 +5,15 @@ Standardize lineage based on GISAID's subtype and lineage.
...
+from pathlib import Path
+import sys
 from typing import Iterable, Dict
 from augur.io.json import dump_ndjson, load_ndjson
-from augur.io.print import print_err
 
+SCRIPT_NAME = Path(sys.argv[0]).stem
+
+def print_err(*args):
+    print(f"[{SCRIPT_NAME}] ", *args, file=sys.stderr)
```
joverlee521: I like this pattern for the curate scripts. I wonder if we should formalize this in Augur as well with the standard logging library.

ingest/rules/curate.smk
```diff
@@ -98,6 +97,7 @@ rule curate:
             | ./scripts/parse-gisaid-location \
                 --location-field {params.gisaid_location_field:q} \
                 --strain-field {params.gisaid_strain_field:q} \
+                --annotations {input.location_annotations:q} \
```
joverlee521: Ah, this has the downside of potentially getting overridden by geolocation rules downstream. I see you what you mean by needing to document the curation steps for geolocations...

ingest/scripts/parse-gisaid-location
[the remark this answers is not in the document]
jameshadfield: I went this way -- seems cleaner in some senses. I don wonder if we still want that Augur change but there's no pressure anymore.

ingest/curation-docs.md
```diff
@@ -0,0 +1,40 @@
...
+
+1. The `parse-gisaid-location` script parses the (capital-L) "Location" field from the GISAID record, which is expected to have slash-separated geographic information.
+  1a. `defaults/locations.tsv` sets custom fields for epi-isls in the TSV. If the record has an annotation then no more modifications are done by the script.
+  1b. `defaults/geolocation_rules.tsv` remaps the slash-separated Location fields. This is mostly used to correct location values with more than 4 fields so that the script can parse them.
+  1c. Sets the slash-separated Location value to "region", "country", "division" and "location"
+  1d. If four fields aren't populated then we attempt to extract a further piece of geographic information from the strain name.
+2. `augur curate titlecase` changes casing
+3. `augur curate apply-geolocation-rules`
+  3a. First runs with Augur default fixes
+  3b. Then runs with our custom rules from `defaults/geolocation_rules.tsv`
```
joverlee521:
```suggestion
  3b. Then runs with our custom rules from `defaults/geolocation_rules.tsv`. This is to ensure that we use specific regions such as "Japan Korea".
```

joverlee521 (approved):
Latest changes look good to me. I think fine to merge and continue improving the curation docs as necessary

Issue #269, opened 2025-12-02
ingest: support multiple of the same segment per record


joverlee521:
_Originally flagged on [Slack](https://bedfordlab.slack.com/archives/C03KWNCLK/p1764713312793569?thread_ts=1764617114.416369&cid=C03KWNCLK)_

Some GISAID records have multiple of the _same_ segment for a single EPI ISL. For example, `EPI_ISL_169351` has two HA segments: `EPI553145` and `EPI553172`.

PR #274, opened 2025-12-08
ingest: Support multiple of the same segment per record

joverlee521:

Within GISAID, a single record can have multiple sequences for the same segment. This commit adds support for multiple sequences of the same segment by changing the expected schema for the "sequences" key in the NDJSON. Then when splitting the NDJSON by segment, it chooses a single sequence per segment by selecting the longest sequence with the latest accession. The latest accession is represented by the biggest integer after the "EPI" prefix in the segment accession. Future improvements can add options for specifying segment accessions and/or sequence quality annotations from Nextclade.

Note I decided to change the schema of the NDJSON to allow us to easily pick the segment sequences from records and easily add future improvements such as deduping strains by available segments. This does mean that we have to reprocess all previous downloads to regenerate the cache, but we needed to do this anyways to include all segment sequences.

## Related issue(s)

Resolves https://github.com/nextstrain/seasonal-flu/issues/269

## Checklist

- [ ] Checks pass
- [ ] Update changelog

### DIFF
--- a/ingest/scripts/diff-ndjson
+++ b/ingest/scripts/diff-ndjson
@@ -105,9 +105,16 @@ def compare_records(truth_record, query_record):
 
 def remove_sequence(record):
     """Remove the sequence string (ATCGs) to save memory"""
-    for s in record.get('sequences', []):
-        if 'sequence' in s:
-            del s['sequence']
+    sequences = record.get('sequences')
+    if isinstance(sequences, list):
+        for s in sequences:
+            if 'sequence' in s:
+                del s['sequence']
+    elif isinstance(sequences, dict):
+        for segment, seqs in sequences.items():
+            for s in seqs:
+                if 'sequence' in s:
+                    del s['sequence']
     return record
 
 if __name__ == '__main__':


ingest/scripts/link-gisaid-metadata-and-fasta
```diff
@@ -12,27 +12,36 @@ Each record represents a single GISAID record, formatted as:
...
-                “segment”: “NA”,
-                “accession”: “...”,
-                “sequence”: “...”
-            },
+        “sequences”: {
+            "ha": [
+                {
+                    “accession”: “...”,
+                    “sequence”: “...”
+                },
```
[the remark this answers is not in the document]
jameshadfield: This should be fine. Most of my work doesn't use the sequences themselves. Places I can see needing updates:


1. `scripts/diff-ndjson` (already in master) attempts to remove the actual sequence contents to save memory. We need to update this function. We may also want to add the capability to flag up when the selected accession changes as part of this script.
1. PR #268 introduces `diff-avian-flu.py` which will need to be changed quite a bit, but these changes should allow us to better match up records which is nice. We can do this later.
1. PR #268 also added a `n_segments` column to the split-gisaid-ndjson-by-segment script. We may want to add that capability here? Else it should rebase without too much trouble.

ingest/scripts/diff-ndjson
```diff
@@ -105,9 +105,16 @@ def compare_records(truth_record, query_record):
...
 def remove_sequence(record):
     """Remove the sequence string (ATCGs) to save memory"""
-    for s in record.get('sequences', []):
-        if 'sequence' in s:
-            del s['sequence']
+    sequences = record.get('sequences')
+    if isinstance(sequences, list):
+        for s in sequences:
+            if 'sequence' in s:
+                del s['sequence']
```
joverlee521: from https://github.com/nextstrain/seasonal-flu/pull/274#pullrequestreview-3565191415:

> Question: record['sequences'] will only be a dict from this point on, but allowing diff-ndjson to handle lists is for backwards compat? If so - thank you!!!
 
Yes, I kept the list support in place for backwards comparison. This let me compare a small dataset new NDJSON against the old NDJSON to make sure the only thing that's changed was the type of the `sequences` field.

ingest/scripts/split-gisaid-ndjson-by-segment
```diff
@@ -159,3 +166,10 @@ if __name__ == '__main__':
         Path(args.output_metadata).touch()
     else:
         write_records_to_tsv(chain([first_record], modified_records), args.output_metadata)
+
+    # Summary statistics (to STDERR)
+    print_err(f"Num records: {sum(STATS.values()):,}")
+    print_err("Num segments per record:")
+    print_err(f"0 segments: {STATS[0]:,} (records not written to metadata)")
+    for i in range(1, 8+1):
+        print_err(f"{i} segments: {STATS[i]:,}")
```
joverlee521: from https://github.com/nextstrain/seasonal-flu/pull/274#pullrequestreview-3565191415:

> Idea: expand the stats in split-gisaid-ndjson-by-segment to also report counts of sequences (for each segment)

I'm not 100% sure what you mean here, so I'll come back to this later.

Issue #234, opened 2025-05-29
Switch builds to use new ingest outputs


joverlee521:
## TODOs

- [x] update geolocation rules 
- [x] remove `/trials/ingest` from S3 prefix in [manual-upload](https://github.com/nextstrain/seasonal-flu/blob/9f268730449282d28d2c51fab2fb1c3e904f1cf6/ingest/build-configs/manual-upload/config.yaml#L3) and [nextstrain-automation](https://github.com/nextstrain/seasonal-flu/blob/9f268730449282d28d2c51fab2fb1c3e904f1cf6/ingest/build-configs/nextstrain-automation/config.yaml#L8-L9)
- [x] update GH Action workflows
    - [x] remove targets `upload_all_raw_sequences`, `upload_all_sequences`, `upload_all_metadata` in [upload job](https://github.com/nextstrain/seasonal-flu/blob/9f268730449282d28d2c51fab2fb1c3e904f1cf6/.github/workflows/upload.yaml#L53-L55)
    - [x] add ingest workflow to upload workflow (before `run-nextclade`)
- [x] copy all files from `s3://nextstrain-data-private/files/workflows/seasonal-flu/trials/ingest/gisaid-downloads/processed/` to prod S3 location
- [x] add docs for curation (might reveal papercuts that we want to tackle before completing switch)
- [ ] save a copy of metadata.tsv and sequences.fasta produced from fauna on S3 (?)
- [ ] add rules/config for reprocessing all data in `nextstrain-automation` (prompted by https://github.com/nextstrain/seasonal-flu/pull/237)

huddlej:
I just ran into a variant of the issue I mentioned above where H1N1pdm titer data include a reference strain named A/SouthAustralia/59/2023. We have this strain in our fauna metadata and sequences, but for the new ingest we only have the strain in the metadata and not in the sequences file.

huddlej:
Thanks for figuring that one out, @joverlee521! I think we will usually want to prioritize records with HA over records without, then records with NA, and then everything else.

joverlee521:
> Thanks for figuring that one out, [@joverlee521](https://github.com/joverlee521)! I think we will usually want to prioritize records with HA over records without, then records with NA, and then everything else.

Thanks @huddlej, moving this into a separate issue that I've linked as a sub-issue.

PR #285, opened 2025-12-19
Switch to ingest in GH Action workflow

jameshadfield:
> [!IMPORTANT]
> This PR should be merged along with https://github.com/nextstrain/seasonal-flu/pull/287 to fully switch to the ingest workflow.

GH Action workflow updates to switch to the new ingest workflow. This keeps `upload` as the main workflow that orchestrates all the moving parts.
1. Allow `ingest` to trigger avian-flu's GenoFLU workflow
2. Make `ingest` a reusable workflow
3. Call `ingest` from within `upload`
4. Trigger avian-flu within `upload` as an independent job that doesn't affect other seasonal-flu jobs

## Related issue(s)

Part of https://github.com/nextstrain/seasonal-flu/issues/234

## Checklist

- [x] [Trial run from upload](https://github.com/nextstrain/seasonal-flu/actions/runs/20444571164)

### DIFF
--- a/.github/workflows/upload.yaml
+++ b/.github/workflows/upload.yaml
@@ -44,10 +44,30 @@ jobs:
       id-token: write
     uses: ./.github/workflows/ingest.yaml
     with:
-      triggerAvianFlu: ${{ inputs.triggerAvianFlu }}
+      # Trigger avian flu in a separate job to prevent error from failing
+      # the rest of the upload workflow
+      triggerAvianFlu: false
       artifact-name: build-outputs-ingest
     secrets: inherit
 
+  trigger_avian_flu:
+    needs: [ingest]
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    steps:
+      - if: ${{ inputs.triggerAvianFlu }}
+        name: Trigger avian-flu/genoflu-gisaid.yaml
+        # TODO XXX - drop --ref once https://github.com/nextstrain/avian-flu/pull/158 is merged
+        # Remove -f once we have switched over to this workflow for avian-flu
+        run: |
+          gh workflow run \
+            genoflu-gisaid.yaml \
+            --repo nextstrain/avian-flu \
+            --ref james/seasonal-flu-ingest \
+            -f trial_name=genoflu-gisaid
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/ingest.yaml
```diff
@@ -28,12 +28,15 @@ jobs:
   ingest:
     permissions:
       id-token: write
+      contents: read
+      actions: write
```
jameshadfield: I don't think these permissions will flow into the reusable workflow as it sets (overwrites) its own permissions

https://github.com/nextstrain/.github/blob/38d02e1e39258a1609c6944bad9413af2462b33b/.github/workflows/pathogen-repo-build.yaml.in#L193-L194

These permissions are (I think) required for the GITHUB_TOKEN to be able to dispatch workflows

joverlee521: Ah, the permissions/secrets/env are much simpler if we orchestrate the trigger completely within the GH Action workflow (e.g. [upload.yaml triggers a bunch of downstream workflows](https://github.com/nextstrain/seasonal-flu/blob/5a94edb61dfdb9249f61ff694781677853bdbe0a/.github/workflows/upload.yaml#L67-L94))

.github/workflows/ingest.yaml
```diff
@@ -28,12 +28,15 @@ jobs:
...
     uses: nextstrain/.github/.github/workflows/pathogen-repo-build.yaml@master
     secrets: inherit
     with:
       runtime: ${{ inputs.runtime }}
       env: |
         NEXTSTRAIN_DOCKER_IMAGE: ${{ inputs.dockerImage }}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
joverlee521: IIRC, the default `GITHUB_TOKEN` within a workflow only has permissions within the repository that triggered it. I believe the `GITHUB_TOKEN` passed in here would not have permissions to trigger workflows in a different repo. This is why we have the org wide `GH_TOKEN_NEXTSTRAIN_BOT_WORKFLOW_DISPATCH` token for dispatching across repos. 

However, as I said in https://github.com/nextstrain/seasonal-flu/pull/285#discussion_r2636305261, I think this orchestration would be a lot simpler if done within the GH Action workflow. I'll push up a separate commit to do this.

PR #291, opened 2026-01-12
Update hardcoded strain names

jameshadfield:
With our move away from fauna and to a curated all-influenza ingest pipeline we have introduced a number of strain name changes. Here we check and correct these strain names where possible, first automatically (commit no. 2) and then manually (commit no. 3). 

The majority of remaining mismatches are cases where the strain name doesn't exist in fauna either.

### DIFF
--- a/config/references_for_titer_plots/h1n1pdm/egg_hi.txt
+++ b/config/references_for_titer_plots/h1n1pdm/egg_hi.txt
@@ -6,19 +6,13 @@ A/Switzerland/6849/2025-egg # D.3.1:113K
 A/California/177/2024-egg # C.1.9.1
 A/Victoria/42/2025-egg # C.1.9.3
 # [update-strains] No match for the following strain, which did exist in fauna:
+# [update-strains] (the non-egg-passaged strain does exist)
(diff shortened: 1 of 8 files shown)


config/references_for_titer_plots/h1n1pdm/egg_hi.txt
```diff
@@ -5,8 +5,12 @@ A/Sydney/124/2025-egg # D.3.1
 A/Switzerland/6849/2025-egg # D.3.1:113K
 A/California/177/2024-egg # C.1.9.1
 A/Victoria/42/2025-egg # C.1.9.3
+# [update-strains] No match for the following strain, which did exist in fauna:
+# [update-strains] (the non-egg-passaged strain does exist)
 A/Norway/10655/2024-egg # C.1.9.3:35N
```
jameshadfield: We get a number of these non-matches across different subtypes, which surprised me a bit. Looking at the two in this file a little more their (fauna) EPI_ISLs are not in the gisaid cache NDJSON. I don't have a good answer for why.

PR #292, opened 2026-01-12
Cross reference titer strains against metadata strains

### LINKED ISSUE #286: Check strain names in titer data
### Context

@jameshadfield recently looked into matches of titer strain names against the strain names of sequences from the new ingest workflow and it's as good/bad as current strain name matches against fauna sequences ([Slack](https://bedfordlab.slack.com/archives/C03KWDET9/p1765500798956549)).

We should add a check to verify that titer strain names match an existing sequence strain name. This would help us identify misspelled strain names such as those in https://github.com/nextstrain/seasonal-flu/issues/60. 

### Description

Even before we move titer ingest out of fauna, we can add this check into the existing workflow for titers.

We currently download titers from fauna, separated by lineage/center/passage/assay, but the `who` center is a superset of all other centers. So after `download_titers`, we can check the `data/<lineage>/who_<passage>_<assay>_titers.tsv` files against the `<lineage>/metadata.tsv`. This would verify the titers' `virus_strain` and `serum_strain` values match `strain` in the metadata. Every value that does not match would get output to a log file.

This would create a path to use the log file to generate a mapping file of bad titer strain names to existing sequence strain names. This can be applied in a curation step before the titer data is uploaded to S3.

jameshadfield:
## Original investigation results

Match titer strain names against fauna (the current status-quo) and against the newly curated metadata. For the latter, do both an exact match and a more complex match, with the idea being that we can add a simple strain-name remapping layer when we download the titers from fauna.

The "more complex" match either uses EPI_ISL to link strain names if it was a fauna match or simplifies (normalizes?) the strain name by removing punctuation, lowercasing etc and compares against similarly simplified metadata strain names. 

### Match rate against all titers

The following plot shows that the match rate for CDC & Crick titers is now almost 100%, especially for recent years. NIID and VIDRL titers match rates see minor improvements but are almost unchanged. Circle size corresponds to number of titers for that year (extracted from the titer strain name). 

titer-matching

### Are we loosing some titers and gaining others?

TL;DR - if we incorporate a thin titer-strain-remapping layer we essentially keep all titers we (fauna) currently match. 

To check we're not seeing big changes in which titers are matched despite the overall percentage improvements (above) we log summary stats as well as verbose per-strain log files. Following is a summary of the summary.

```
subtype   contrib year      will_lose(simple_match)  will_gain(simple_match)  will_lose(after_remap)   will_gain(after_remap)   
h3n2      cdc     ALL       9                        180                      4                        191                      
h3n2      vidrl   ALL       21                       410                      2                        736                      
h3n2      niid    ALL       4                        30                       4                        32                       
h3n2      crick   ALL       27                       210                      3                        232                      
h3n2      ALL     ALL       61                       830                      13                       1191                     
h1n1pdm   vidrl   ALL       4                        80                       0                        156                      
h1n1pdm   niid    ALL       1                        8                        0                        8                        
h1n1pdm   cdc     ALL       9                        272                      3                        276                      
h1n1pdm   crick   ALL       27                       157                      2                        186                      
h1n1pdm   ALL     ALL       41                       517                      5                        626                      
vic       vidrl   ALL       9                        32                       0                        66                       
vic       niid    ALL       0                        8                        0                        9                        
vic       cdc     ALL       5                        223                      0                        237                      
vic       crick   ALL       19                       77                       0                        91                       
vic       ALL     ALL       33                       340                      0                        403                      
yam       vidrl   ALL       3                        35                       0                        195                      
yam       niid    ALL       1                        1                        0                        1                        
yam       cdc     ALL       4                        350                      0                        351                      
yam       crick   ALL       2                        70                       0                        92                       
yam       ALL     ALL       10                       456                      0                        639 
```


scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,231 @@
...
+import gzip
+import lzma
+import io
+import csv
+import json
+import re
+from augur.io import read_metadata
+from typing import Any
+
+def open_file(file_path: str):
```
joverlee521: This can be replaced with `augur.io.open_file` so we don't have to maintain decompression here.

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -86,6 +86,36 @@ rule download_sequences:
...
+    params:
+        s3_path=S3_PATH + "/{lineage}/metadata.tsv.xz"
+    conda: "../../workflow/envs/nextstrain.yaml"
+    shell:
+        """
+        aws s3 cp {params.s3_path} - > {output.metadata}
+        """
+
+# Similarly this rule downloads from S3
+rule download_fauna_strain_map:
```
joverlee521: Is the `fauna-strain-map.tsv.xz` on S3 a manually generated file from the previous fauna metadata? I assume it will not need to be updated?

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -111,6 +141,30 @@ rule download_titers:
             --fstem {wildcards.lineage}/{wildcards.center}_{wildcards.passage}_{wildcards.assay} 2>&1 | tee {log}
         """
 
+rule remap_titer_strain_names:
+    input:
+        # NOTE: this script assumes the {lineage}/{center}_{passage}_{assay} file name structure!
+        titers = "data/{lineage}/{center}_{passage}_{assay}_titers.tsv",
```
joverlee521: If worried about breakage from file name structure changing, I'd push for providing the wildcard values to the script directly.

joverlee521 (commented):
The Snakemake workflow changes make sense to me! Also reviewed the `remap-titer-strain-names` script since I was here.

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,231 @@
...
+    matches: dict[str, Any] = {
+        'unique_strains': {},
+        'measurements': {},
+    }
+
+    fauna_map = read_fauna_map(args.fauna_strain_map)
+    titers = read_titers(args.titers)
+    metadata = parse_metadata(args.metadata)
+    epi_isl_to_strain = {v[0]:k for k,v in metadata.items()}
+    simplified_metadata = {simplify_strain(k): k for k in metadata.keys()}
```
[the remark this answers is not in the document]
jameshadfield: Good idea - I'll check and only consider unique simplified (curated metadata) strain names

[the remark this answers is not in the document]
jameshadfield: Done - see [slack conversation](https://bedfordlab.slack.com/archives/C03KWDET9/p1768340617278469). We now log the output and track "maybe" matches for viz purposes, but we don't actually change such titer strain names.

jameshadfield:
### Workflow changes

As part of the `upload_titers` workflow we now:
1. Download curated metadata
1. Download a lookup table of fauna strain name → EPI_ISL which was created when sequences were still stored in fauna (i.e. this file will not change)
1. Update `virus_strain` and `serum_strain` names (in the titers TSVs, as downloaded from fauna) if we are able to find a match in the curated metadata. See below for how we do this. This remapped TSV has two added columns: `virus_strain_match` and `serum_strain_match`, with values of "yes", "maybe" or "no". We use this remapped TSV for downstream rules. 
1. Produce per-lineage visualization for how well our titers are able to be matched against our metadata. These are uploaded to S3.

### Running locally

If you want to run locally you can use a development-rule to run the upload workflow without actually uploading anything! `snakemake --cores 4 -pf dev_only_all_titers --configfile profiles/upload.yaml` 

### Strain name matching approach

* If the strain name is in the hardcoded mapping TSV _and_ the curated metadata, then we use the hardcoded name. (Otherwise we use the uncorrected name.)
* If the strain name is in the metadata we use it unchanged.
* If the titer name was found in the fauna strain → EPI_ISL (step 2 above), then we use this to remap the strain name if possible¹.
* We identify plausible ("maybe") matches by comparing simplified versions² of the strain names; we log such matches as potential fixes, but don't actually change the strain names in the titers TSV.


### Visualisation 

We visualize the matching results (one figure per lineage) and upload PNGs to S3. The visualisation is in small-multiples of center, passage and assay, but could be further split by host if required.

### Final thoughts

I don't have a great answer for why specific titer strains don't match.

---

¹ Not all of them, but nearly all of them. Since we use EPI_ISL as the merge key, if the fauna sample's EPI_ISL wasn't in our curated metadata then we can't match it. 

² Simplification means: lowercase, punctuation removed, leading zeros removed from numbers, and some other ad-hoc fixes.

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,304 @@
...
+
+    if titer_strain in fauna_map:
+        epi_isl = fauna_map[titer_strain]
+        if epi_isl in epi_isl_to_strain:
+            curated_strain = epi_isl_to_strain[epi_isl]
+            cache[titer_strain] = curated_strain
+            return ("yes", curated_strain)
+    
+    # TODO XXX add a way to provide a hardcoded mapping of titer-strain-name to new-strain-name
+    # and apply those matches here.
```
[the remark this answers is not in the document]
joverlee521: Thanks for [flagging this PR](https://github.com/nextstrain/private/issues/210#issuecomment-4093355489). I still think it'd be worth it to implement the escape hatch for titer strain name matching as part of this PR.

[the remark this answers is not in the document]
jameshadfield: Now implemented.

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,304 @@
...
+    and the year they are from.
+    We track (unique) individual strains separately to (unique) measurements (pairs of strains);
+    for the latter, the year is taken as the virus_strain not the serum_strain.
+    """
+    virus_strain_year = get_year(virus_match[1], metadata) if virus_match[1] else get_year(original_virus_strain)
+    serum_strain_year = get_year(serum_match[1], metadata) if serum_match[1] else get_year(original_serum_strain)
+
+    # store whether the individual virus & serum strains matched:
+    matches['strains'][original_virus_strain] = (virus_match[0], virus_strain_year)
+    matches['strains'][original_serum_strain] = (serum_match[1], serum_strain_year)
```
joverlee521: This should be storing the yes/no/maybe value, which is the first value in the Match tuple right?

```suggestion
    matches['strains'][original_virus_strain] = (virus_match[0], virus_strain_year)
    matches['strains'][original_serum_strain] = (serum_match[0], serum_strain_year)
```

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # and store whther the pair of strains (i.e. the measurement) matched
+    if virus_match[0]=='no' or virus_match[0]=='no':
+        measurement_match_result = 'no'
+    elif virus_match[0]=='maybe' or virus_match[0]=='maybe':
+        measurement_match_result = 'maybe'
```
joverlee521: I think this should be checking `virus_match` or `serum_match`?

```suggestion
    if virus_match[0]=='no' or serum_match[0]=='no':
        measurement_match_result = 'no'
    elif virus_match[0]=='maybe' or serum_match[0]=='maybe':
        measurement_match_result = 'maybe'
```

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,334 @@
...
+    return strain_map
+
+UNKNOWN_YEAR = 'unknown'
+def parse_year(date:str) -> int|str:
+    try:
+        return int(date[0:4])
+    except ValueError:
+        return UNKNOWN_YEAR
+
+type Metadata = dict[str, tuple[str, int|str]]
```
huddlej: [The `type` keyword first appears in Python 3.12](https://docs.python.org/3/library/typing.html#type-aliases), but the Nextstrain Docker image still uses 3.11. The following is backward compatible.

```suggestion
# Also requires imports at top to include:
# from typing import TypeAlias
Metadata: TypeAlias = dict[str, tuple[str, int|str]]
```

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,334 @@
...
+        else:
+            dropped.append(true_strains)
+    if dropped:
+        count = sum([len(x) for x in dropped])
+        print(f"[WARNING] In total we dropped {count:,} curated strains for simplified matching due to collisions once simplified", file=sys.stderr)
+        # Comment out the following if you want to see them all, but I don't want to clutter the automated logs
+        # print('[WARNING] Dropped strains: ' + ", ".join(["{" + ', '.join([f"{s!r}" for s in strains]) + "}" for strains in dropped]), file=sys.stderr)
+    return simplified
+
+type Match = tuple[str, str|None]
```
huddlej:
```suggestion
Match: TypeAlias = tuple[str, str|None]
```

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,304 @@
...
+    by_year['total'] = [0,0,0]
+    for match, year in matches.values():
+        if match=='yes':
+            by_year[str(year)][0] += 1
+            by_year['total'][0] += 1
+        elif match=='maybe':
+            by_year[str(year)][1] += 1
+            by_year['total'][1] += 1
+        by_year[str(year)][2] += 1
+        by_year['total'][2] += 1
```
joverlee521: _non-blocking comment_

In general, I'm wary of using lists/tuples to track data like this because it's prone to errors when accessing by index and also it's not very readable of what data we're tracking.

profiles/upload.yaml
```diff
@@ -71,6 +71,11 @@ recency:
   date_bin_labels: ["last week", "last month", "last quarter"]
   upper_bin_label: older
 
+titer_strain_maps:
+  h1n1pdm: "config/h1n1pdm/titer-strain-map.tsv"
+  h3n2: "config/h3n2/titer-strain-map.tsv"
+  vic: "config/vic/titer-strain-map.tsv"
```
joverlee521, line 78: _non-blocking_

Seeing this made me wonder about Yam...It's not included in this config because we no longer receive Yam titers, but I wonder if we should update it with the remapped strain names.

joverlee521 (changes requested):
Thanks for flagging this again, agree that we should get this in as a first step to smoothing out titer ingest.

config/h1n1pdm/titer-strain-map.tsv
```diff
@@ -0,0 +1 @@
+FAUNA_TITER_STRAIN	CURATED_STRAIN
```
huddlej: Do we plan to use these files in a post-fauna future to handle other strain name mismatches between the titer source of truth and the sequence source of truth? If so, we might rename the first column to just `TITER_STRAIN` in these strain map files.

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -86,6 +88,38 @@ rule download_sequences:
             --fstem {wildcards.lineage}/raw_{wildcards.segment} 2>&1 | tee {log}
         """
 
+S3_PATH = config.get("s3_path", "s3://nextstrain-data-private/files/workflows/seasonal-flu")
```
huddlej: S3 path is already defined for this Snakefile through the `config['s3_dst']` variable that gets [set in the upload.yaml file](https://github.com/nextstrain/seasonal-flu/blob/b300faec0394d7baaaece1246e33c7d090c15494/profiles/upload.yaml#L4). The `S3_PATH` in `download_from_s3.smk` is an unnecessary reimplementation of this config-based approach. We could reasonably move this config to the `Snakefile` and let users override it as they like.

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -86,6 +88,38 @@ rule download_sequences:
...
+
+# This is almost identical to `download_parsed_metadata` in `download_from_s3.smk` however
+# we include it here so that fauna (titer) workflows can access it. If we import the
+# rules file we get a number of rule name & file name clashes which I (James) don't
```
huddlej: It seems better to not copy/paste the rule for downloading metadata, when we can include it from another rule set. I don't mind fixing those name clashes. We can delete `download_sequence` from the fauna rules and handle other clashes with `ruleorder` declarations. I can push a patch for this later.

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -135,3 +197,34 @@ rule select_titers_by_host:
         """
         tsv-filter -H {params.host_query} {input.titers} > {output.titers} 2> {log}
         """
+
+
+def _stats(wildcards):
```
huddlej:
```suggestion
def _get_titer_matching_stats(wildcards):
```

workflow/snakemake_rules/download_from_fauna.smk
```diff
@@ -135,3 +197,34 @@ rule select_titers_by_host:
...
+    return sorted(list(stats))
+
+rule visualise_titer_matches:
+    """
+    Visualise the titer matche stats produced by the remap_titer_strain_names
+    rule above. This script will _always_ exit 0 and produce the output file
+    to avoid viz errors terminating the pipeline
+    """
+    input:
+        stats = _stats
```
huddlej:
```suggestion
        stats = _get_titer_matching_stats
```

huddlej (commented):
Thank you for doing all the hard bookkeeping work here, @jameshadfield! We should have done this years ago...

The only blocking comments below relate to Python syntax issues for 3.11 vs. 3.12. Others are minor. I can push a patch tomorrow morning for my suggestion below to not copy/paste the metadata download rule, but it's not a big deal either way.

The output of these new rules/scripts focuses primarily on the number of _matches_, but I'm most interested in the list of strain names without matches. In the case of VIDRL data, we know that they run serological assays before they sequence, so it isn't surprising to find the greatest number of measurements that exist without matching sequences. The cases of the other CCs should be more interesting to track down. For example, we have CDC measurements for H1 strain A/Minnesota/115/2024, but that strain doesn't exist in the sequence database nor does it exist in the current CDC records (serum strain: A/Alaska/14/2023, ferret id: 2024-020). Where did that measurement come from? If it got redacted by CDC for whatever reason, we should redact it too.

profiles/upload.yaml
jameshadfield: Right - Yam is not here because [it's not part of the upload profile](https://github.com/nextstrain/seasonal-flu/blob/master/profiles/upload.yaml) and this titer matching approach operates only in the context of that workflow. I'd think that adding it would be better as a separate PR if you / John want to follow up here?

scripts/remap-titer-strain-names.py
jameshadfield: Bah! Will fix

jameshadfield:
> The output of these new rules/scripts focuses primarily on the number of matches, but I'm most interested in the list of strain names without matches.

I'll tidy up all the fixes tomorrow and also produce an output of the _missing_ titer strains which can be followed up manually. Thanks for the dual reviews!

workflow/snakemake_rules/download_from_fauna.smk
huddlej: @jameshadfield I pushed a patch in [6688f28](https://github.com/nextstrain/seasonal-flu/pull/292/commits/6688f28e4a5dcb7a7535950e471bff494896764b) to resolve naming conflicts and allow reuse of the existing metadata download rule. You can drop this, if you find the rule duplication simpler than the solution I've proposed, though.

jameshadfield: Thanks! Included (although commit sha changed)

workflow/snakemake_rules/download_from_fauna.smk
jameshadfield: Resolved by https://github.com/nextstrain/seasonal-flu/pull/292#discussion_r2978196450

config/h1n1pdm/titer-strain-map.tsv
jameshadfield: Yeah I think we do - updated the name

scripts/remap-titer-strain-names.py
jameshadfield: Urgh supporting type hints over different python versions is such a pain it almost makes me not bother using them in the first place. Thanks for the fixes -- both applied, although with the changes I've made I've probably introduced some more :(

scripts/remap-titer-strain-names.py
jameshadfield: I've switched to using [Data Classes](https://docs.python.org/3/library/dataclasses.html), one of the ~thousand ways python allows these kinds of structures. It makes the code more verbose but a lot more readable.

scripts/remap-titer-strain-names.py
jameshadfield: Now using Data Classes - see https://github.com/nextstrain/seasonal-flu/pull/292#discussion_r2985065480

scripts/remap-titer-strain-names.py
jameshadfield: Refactored things a lot with the introduction of Data Classes so this code is no longer

jameshadfield:
I've made significant changes to the code structure (using Data Classes for much more code clarity). I'm much happer with the readability. There were some slight changes in the results which are due to the bugs @joverlee521 identified in review. Here's an example of the changes:

```diff
- Total matching (unique) strains:    3,976/4,012 (99.1%)
- Total matching (unique) measurement pairs: 8,067/8,078 (99.9%)
- Potential matching (unique) strains:    1/4,012 (0.0%)
- Potential matching (unique) measurement pairs: 5/8,078 (0.1%)
+ Total matching (unique) strains:    4,008/4,012 (99.9%)
+ Total matching (unique) measurement pairs: 8,041/8,078 (99.5%)
+ Potential matching (unique) strains:    2/4,012 (0.0%)
+ Potential matching (unique) measurement pairs: 31/8,078 (0.4%)
```

@huddlej I've also added a commit to report unmatched titer strain names across all (non-yam!) titers TSVs and upload this too. Here's a few examples of what that TSV looks like:

```
titer_strain                               virus_strain_count   serum_strain_count   potential_matching_strain             lineages       centers     passages   assays
----------------------------------------   ------------------   ------------------   -----------------------------------   ------------   ---------   --------   ------
A/Minnesota/115/2024                       16                   0                                                          h1n1pdm        cdc,who     cell,egg   hi
A/India/PUN-NIV364610/2021                 28                   0                    A/India/Pun-NIV364610/2021            h1n1pdm        cdc,who     cell,egg   hi
A/Astrak/1/2011                            0                    424                                                        h1n1pdm        crick,who   cell       hi
A/LvivN/6/2009                             0                    392                  A/Lviv/N6/2009                        h1n1pdm        crick,who   cell       hi
```

profiles/upload.yaml
joverlee521: We can follow up in https://github.com/nextstrain/seasonal-flu/issues/304.

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,350 @@
...
+        count = sum([len(x) for x in dropped])
+        print(f"[SIMPLIFIED MATCHING] When creating a lookup of simplified strain name → actual strain name, we dropped {count:,} strains due to collisions once simplified", file=sys.stderr)
+        # Comment out the following if you want to see them all, but I don't want to clutter the automated logs
+        # print('[WARNING] Dropped strains: ' + ", ".join(["{" + ', '.join([f"{s!r}" for s in strains]) + "}" for strains in dropped]), file=sys.stderr)
+    return simplified
+
+@dataclasses.dataclass
+class Match:
+    in_metadata: bool # is the (new) strain name found in metadata?    
+    name_change: bool # is the (new) strain_name different?
```
joverlee521: _non-blocking nit_

I don't think `name_change` is used anywhere? Can it be removed or be converted to a property attribute so it doesn't get out of sync with `original_name` and `new_name`? 

```python
@property 
def name_change(self) -> bool: 
    return self.original_name != self.new_name
```

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,403 @@
...
+                counts[virus_match.original_name].potential_match = str(virus_match.new_name_matching_when_simplified)
+    
+        if not serum_match.in_metadata:
+            counts[serum_match.original_name].n_serum_strain += 1
+            if serum_match.new_name_matching_when_simplified:
+                counts[serum_match.original_name].potential_match = str(serum_match.new_name_matching_when_simplified)
+    
+    # hardcoded header information as we expect certain information in --stats-metadata
+    header = ['titer_strain', 'virus_strain_count', 'serum_strain_count', 'potential_matching_strain',
+        'lineages', 'centers', 'passages', 'assays']
```
jameshadfield: @huddlej maybe `year` would be useful information for this table?

scripts/remap-titer-strain-names.py
```diff
@@ -0,0 +1,350 @@
...
+            dropped.append(true_strains)
+    if dropped:
+        count = sum([len(x) for x in dropped])
+        print(f"[SIMPLIFIED MATCHING] When creating a lookup of simplified strain name → actual strain name, we dropped {count:,} strains due to collisions once simplified", file=sys.stderr)
+        # Comment out the following if you want to see them all, but I don't want to clutter the automated logs
+        # print('[WARNING] Dropped strains: ' + ", ".join(["{" + ', '.join([f"{s!r}" for s in strains]) + "}" for strains in dropped]), file=sys.stderr)
+    return simplified
+
+@dataclasses.dataclass
+class Match:
```
joverlee521: Should these be read-only instances to prevent unintentional changes? 

```suggestion
@dataclasses.dataclass(frozen=True)
class Match:
```

joverlee521 (approved):
Thanks for continuing to improve the process here! Only left non-blocking nitpicky comments, otherwise looks good to me.

scripts/remap-titer-strain-names.py
jameshadfield: Added year and sorting in force-push. We now order chronologically and, within a year, by the number of virus + serum observations

scripts/remap-titer-strain-names.py
jameshadfield: Yeah, I didn't end up using this value in the end. I've removed it in force push. Thanks for checking!

scripts/remap-titer-strain-names.py
jameshadfield: Yeah, good idea. Done in force push :)

jameshadfield:
PR closes #304

huddlej:
> I've also added a commit to report unmatched titer strain names across all (non-yam!) titers TSVs

This excellent! Thanks again, @jameshadfield!

PR #294, opened 2026-01-15
ingest: Prioritize records by segment

### LINKED ISSUE #253: ingest: prioritize records by available segments
Originally from https://github.com/nextstrain/seasonal-flu/issues/234#issuecomment-3304447242

> I think we will usually want to prioritize records with HA over records without, then records with NA, and then everything else.

This would probably involve updating the [dedup-by-strain script](https://github.com/nextstrain/seasonal-flu/blob/8669db1b8e4825fbe6faf48a46d60a1452f1e964/ingest/scripts/dedup-by-strain). The brute force method would be to loop through the records twice. First to find the prioritized record per strain and second to output those prioritized records.

joverlee521:

For records that have the same strain name, prioritize records that have HA segments and then records that have NA segments. Otherwise, keep the first record in the NDJSON.

## Related issue(s)

Resolves https://github.com/nextstrain/seasonal-flu/issues/253

## Checklist

- [x] Checks pass
- [ ] Update changelog


ingest/scripts/prioritize_id_per_strain
```diff
@@ -0,0 +1,176 @@
...
+            if use_new_record(prioritized["reasons"], new_reasons):
+                prioritized_ids[strain] = {
+                    "id": record_id,
+                    "reasons": new_reasons
+                }
+        else:
+            prioritized_ids[strain] = {
+                "id": record_id,
+                "reasons": get_reasons(sequences, first_record = True)
+            }
```
jameshadfield: When reading this I wondered about a slightly different design. Instead of keeping track of a single "prioritised" ID per strain, for each strain store a list of IDs and their `reasons` . You can then do a second pass through this to make a descision about the ID to pick. E.g.:
* Clear winner? (e.g. hardcoded, only one with HA, only one with HA + NA etc)
* Multiple potential winners? (e.g. multiple with both HA & NA)

We could start by logging the multiple potential winners ("unresolved duplicates"), and over time expand our decision algorithm to incorporate num segments sequenced, date submitted etc. At the least this would be interesting information to log. 

P.S. This would allow you to write out a list of EPI_ISLs to take and then replace `./scripts/dedup-by-strain` with a simple filter.

joverlee521: I didn't get to this, happy to continue this when I get back. If the need arises for prioritizing by segments for flu reports, I think this is also okay to merge as-is.

joverlee521: Coming back to this, I decided to go with your suggestion of a 2 step process: 
1. prioritize records per strain and output prioritized ids
2. filter NDJSON by ids

This now also includes a log of all the strains and records with prioritization reasons. I think this is now more readable of what's happening.

ingest/scripts/filter-ndjson-by-value
```diff
@@ -0,0 +1,41 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+
+    parser.add_argument("--field", required=True,
+        help="The record field for filtering.")
+    parser.add_argument("--include", required=True,
+        help="A text file with values to include, with one value per line.")
+
+    args = parser.parse_args()
+
+    include_values = read_entries(args.include)
```
jameshadfield: Let's convert `include_values: list[str]` to a set, so that the `record[field] in include_values` becomes faster. I remember debugging a augur / treetime script a while back and this exact speedup sped things up by multiple orders of magnitude, although it depends on the list size of course.

jameshadfield (changes requested):
Didn't test, but from reading the code this looks great (apart from one minor change request)

ingest/scripts/filter-ndjson-by-value
joverlee521: Good point! Updated to set in [force-push](https://github.com/nextstrain/seasonal-flu/compare/db194b8dac3d9fca953146d4fb2381b4057f8d0e..c53184629a546c2b39ea08184c5a8af87d4ec1bd)

jameshadfield (approved):

joverlee521:
Tested workflow locally and changes in records were due to prioritizations of HA and NA segments. HA is always _added_ whereas NA is sometimes removed due to prioritization of a record with the HA segment. 

<details>
<summary> HA and NA changes per lineage </summary>

Used `daff diff` to compare metadata.tsv. Viewed output with `visidata` to filter to changes and group by `ha` and `na` columns.

### H3N2 = 1519 changes

|ha           |na           |count  |
|-------------|-------------|------:|
|True         |False\-\>True|1153   |
|False\-\>True|True\-\>False|194    |
|False\-\>True|False\-\>True|98     |
|False\-\>True|True         |45     |
|False\-\>True|False        |28     |
|False        |False\-\>True|1      |

### H1N1pdm = 718 changes

|ha           |na           |count  |
|-------------|-------------|------:|
|True         |False\-\>True|313    |
|False\-\>True|False\-\>True|188    |
|False\-\>True|True         |140    |
|False\-\>True|True\-\>False|69     |
|False\-\>True|False        |8      |

### Vic = 730 changes

|ha           |na           |count  |
|-------------|-------------|------:|
|True         |False\-\>True|607    |
|False\-\>True|False        |45     |
|False\-\>True|True\-\>False|45     |
|False\-\>True|True         |16     |
|False        |False\-\>True|9      |
|False\-\>True|False\-\>True|8      |

### Yam = 390 changes

|ha           |na           |count  |
|-------------|-------------|------:|
|True         |False\-\>True|311    |
|False\-\>True|False\-\>True|41     |
|False\-\>True|True\-\>False|24     |
|False\-\>True|True         |13     |
|False\-\>True|False        |1      |

### Avian-flu = 379 changes

|ha           |na           |count  |
|-------------|-------------|------:|
|True         |False\-\>True|118    |
|False\-\>True|False\-\>True|111    |
|False\-\>True|True\-\>False|85     |
|False\-\>True|True         |36     |
|False\-\>True|False        |26     |
|False        |False\-\>True|3      |

</details>

---

I'm pretty confident this is working as intended, so I'll merge this before ingest on Thursday if there's no other feedback.

PR #298, opened 2026-01-24
Reduce threads

rneher:
- the `profiles/nextflu-private.yaml` profile only runs 3 trees. 8 cores are enough (it really only needs 3)
- For others, make sure there are 4GB per core
- Adjust refine mem to 4GB, others to 2GB (inconsequential since there is 4GB per core).

### DIFF
--- a/.github/workflows/run-nextflu-private-builds.yaml
+++ b/.github/workflows/run-nextflu-private-builds.yaml
@@ -21,8 +21,8 @@ jobs:
       run: |
         nextstrain build \
           --detach \
-          --cpus 16 \
-          --memory 72gib \
+          --cpus 8 \
+          --memory 32gib \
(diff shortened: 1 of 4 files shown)


joverlee521 (approved):
Thanks for digging into this @rneher! Happy to merge this and reduce runtime + costs all around 🎉

trvrb:
Further updated here https://github.com/nextstrain/seasonal-flu/commit/fbb4731cec80ceaf0098f34fbc2b6475a0c32aab

This was primarily to switch to more efficient c7a instances from c5 instances. Reducing from 36 CPUs to 32 CPUs to match c7a instances instead of c5 instances, where c7a.8xlarge has 32 CPUs while c5.9xlarge had 36 CPUs.

PR #306, opened 2026-03-25
[titer strain curation] add hardcoded pattern matching

jameshadfield:
Which is applied after other matching approaches, and only if the resulting titer strain is found in the metadata TSV. Patterns obtained by inspecting `data/missing-titer-strains.tsv` for those strains which had a potential (simplified) match, and then manually checking each one.

The matching improvements are variable - many datasets see a <2% improvement in both strain & measurement matches. However some are much larger, e.g. h3n2/vidrl/egg/fra matched strains: 84% → 95%, measurements 84% → 93%

See <https://bedfordlab.slack.com/archives/C03KWDET9/p1774472953494449> for discussion about this direction and future plans.

### DIFF
--- a/scripts/remap-titer-strain-names.py
+++ b/scripts/remap-titer-strain-names.py
@@ -18,6 +18,45 @@ import dataclasses
 from datetime import date
 import re
 
+HARDCODED_CASE_CHANGES: dict[str,str] = {
+    # (incorrect) titer strain substring, corrected strain substring
+    "/Americansamoa/":  "/AmericanSamoa/",
+    "/Angthong/":  "/AngThong/",
+    "/AnNahdah/":  "/An_Nahdah/",
+    "/BayofPlenty/":  "/BayOfPlenty/",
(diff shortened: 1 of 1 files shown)


scripts/remap-titer-strain-names.py
```diff
@@ -18,6 +18,45 @@
 from datetime import date
 import re
 
+HARDCODED_CASE_CHANGES: dict[str,str] = {
```
joverlee521, line 21: _For future reference_

If we decide to use the standardize-strain-name script for titer ingest in the future, these can easily be moved to the [location-replacements TSV](https://github.com/nextstrain/seasonal-flu/blob/219e21a41952578fdde06206436564abf72a4ac0/ingest/scripts/standardize-strain-names#L347-L349) used by the script.

joverlee521 (approved):
Nice! Good to get these corrected for now and should be an easy move to the curation script in the future.

PR #307, opened 2026-04-01
General clean up after ingest migration

joverlee521:

Clean up config and workflow rules that were used specifically for processing sequences from fauna. 
The major change is the default workflow now starts from the `metadata.tsv` and `<segment>.fasta` instead of the `raw_<segment>.fasta`, see commits for more details.

## Checklist

- [x] Checks pass
- [x] Update changelog


jameshadfield (approved):
> I think it's worth reducing the complexity of the workflow by removing the less used input route.

💯 👍 

I haven't tested or inspected the DAG changes, but from reading the commit messages & the code this looks great

huddlej (approved):
Looks great!

PR #310, opened 2026-04-09
Ingest with Nextclade and haplotype annotations

### LINKED ISSUE #288: Annotate Nextclade TSV with emerging haplotypes and amino acid haplotypes during ingest
### Context

We have at least two workflows (here and forecasts-flu) that download Nextclade TSVs and annotate emerging haplotypes and amino acid haplotypes in a new copy of each TSV.

### Description

@joverlee521 [points out that we could annotate these columns to the Nextclade TSV once during ingest](https://github.com/nextstrain/forecasts-flu/pull/26#discussion_r2644248396) from this repo, avoiding the redundant complexity in our workflows. This is probably the right place to end up. This approach would require us to remember to run the Nextclade workflow when we update the emerging haplotypes TSV in this repo or be content with waiting for our weekly ingest process to rerun the Nextclade workflow.

joverlee521:

Add Nextclade and haplotype annotations to the ingest output. These are optionally included for datasets by defining the `nextclade` config parameters. 

## Related issue(s)

Resolves https://github.com/nextstrain/seasonal-flu/issues/288

## Checklist

- [x] Checks pass
- [x] [set threads for Nextclade](https://github.com/nextstrain/seasonal-flu/pull/310/commits/a3dc76b499b6f7f3d64d3703b2d247738aecf297#r3061276320)
- [x] [Confirm files to upload to S3](https://github.com/nextstrain/seasonal-flu/pull/310/files#r3061286327)
- [x] [test with GISAID download](https://bedfordlab.slack.com/archives/C03KWDET9/p1776126851656569?thread_ts=1775078611.483489&cid=C03KWDET9) (Thanks @huddlej!)
- [x] Remove [run-nextclade](https://github.com/nextstrain/seasonal-flu/blob/d95f3212b0d5f1894bc39eef3ca2725b619c1a49/.github/workflows/run-nextclade.yaml) and remove from [upload](https://github.com/nextstrain/seasonal-flu/blob/d95f3212b0d5f1894bc39eef3ca2725b619c1a49/.github/workflows/upload.yaml#L67-L74) GH Action workflow
- [x] test in GH Actions ([docker](https://github.com/nextstrain/seasonal-flu/actions/runs/24748218887) - completed in 41minutes)
- [x] Update changelog


ingest/defaults/config.yaml
```diff
@@ -238,3 +238,53 @@ filtering:
...
+    ha:
+      dataset_name: nextstrain/flu/h3n2/ha
+      # Field used to merge with the metdata.
+      id_field: seqName
+      # Used to rename fields and only listed fields will be merged with the
+      # metadata in final output. Must include the id_field defined above.
+      field_map:
+        seqName: seqName
+        clade: subclade_nextclade_ha
+        qc.overallStatus: qc.overallStatus
```
joverlee521, line 272: Open question: What other columns should be included from Nextclade?

ingest/rules/nextclade.smk
```diff
@@ -0,0 +1,138 @@
...
+        "logs/run_nextclade/{dataset}/{segment}.txt"
+    shell:
+        r"""
+        exec &> >(tee {log:q})
+
+        nextclade3 run \
+            {input.sequences:q} \
+            --input-dataset {input.dataset:q} \
+            --output-tsv {output.nextclade:q}
+        """
```
joverlee521, line 52: Reminder for myself: add threads. Current workflow [uses 8](https://github.com/nextstrain/seasonal-flu/blob/d95f3212b0d5f1894bc39eef3ca2725b619c1a49/workflow/snakemake_rules/select_strains.smk#L143), but the GH Action workflow [sets it to 12](https://github.com/nextstrain/seasonal-flu/blob/d95f3212b0d5f1894bc39eef3ca2725b619c1a49/.github/workflows/run-nextclade.yaml#L48). Potentially set to 1 if this can just run in the docker runtime in GH Actions.

ingest/defaults/config.yaml
```diff
@@ -238,3 +238,53 @@ filtering:
   #   additional_field: host
   #   additional_field_values: human
 
+# Used to determine which dataset and segments will include Nextclade annotations
+# First level of keys are dataset names, they should matching the `filtering` keys above.
+# Second levele of keys are segments, they will only get used if the segment is
+# also included in the top level `segments` config.
+nextclade:
```
joverlee521, line 245: Open question: What other Nextclade datasets do we want to run? Should we be running all segments? What about yam, b, h1n1, h2n2?

ingest/rules/nextclade.smk
```diff
@@ -0,0 +1,231 @@
...
+            --verbose
+        """
+
+
+rule run_nextclade:
+    input:
+        dataset="data/nextclade/{dataset}/{segment}.zip",
+        sequences="results/{dataset}/{segment}.fasta",
+    output:
+        nextclade="results/{dataset}/{segment}/nextclade.tsv",
```
joverlee521, line 37: Open question: Should this nextclade.tsv be separately uploaded to S3 even though we are merging a subset of columns with the metadata.tsv?

ingest/rules/nextclade.smk
```diff
@@ -0,0 +1,231 @@
...
+            field_map[derived_haplotype_field] = derived_haplotype_field
+
+    return field_map
+
+
+rule nextclade_metadata:
+    input:
+        nextclade=_get_nextclade,
+    output:
+        nextclade_metadata=temp("results/{dataset}/{segment}/nextclade_metadata.tsv"),
```
huddlej: I can myself getting confused by the "metadata" in this file name when elsewhere that term refers to metadata for sequences. What if this intermediate file was named like `subset_nextclade.tsv` like the `subset_metadata.tsv` elsewhere?

ingest/rules/nextclade.smk
```diff
@@ -0,0 +1,231 @@
...
+    params:
+        clade_column=lambda w: _get_nextclade_config(w)["emerging_haplotypes"]["clade_column"],
+        emerging_haplotype_column=lambda w: _get_nextclade_config(w)["emerging_haplotypes"]["emerging_haplotype_column"],
+    benchmark:
+        "benchmarks/{dataset}/{segment}/add_emerging_haplotypes.txt"
+    log:
+        "logs/{dataset}/{segment}/add_emerging_haplotypes.txt"
+    shell:
+        r"""
+        exec &> >(tee {log:q})
```
huddlej, line 103: I hadn't noticed this pattern for teeing logs before. Is the idea for this approach to avoid accidental issues with editing the final line of the command to have the `tee` at the end or is the goal more to announce how the logging is happen more clearly outside of the context of the command that follows?

ingest/defaults/config.yaml
huddlej: This is probably a question for @rneher, but I'd guess it would be nice to have current Nextclade annotations for any extant flu lineages and segments that we have Nextclade datasets for.

ingest/rules/nextclade.smk
huddlej: I started uploading the Nextclade TSV mainly to grab the clade and QC details, but it's handy to have that TSV available for a quick look at any specific sequence. There is so much more in the full TSV that unless we include all columns with the metadata, we probably want the Nextclade TSV, too.

ingest/defaults/config.yaml
huddlej: We probably want `qc.overallStatus` for NA, too, so we can drop low-quality NA sequences before any analyses.

ingest/rules/nextclade.smk
joverlee521: This is following our latest [Snakemake guideline for logs](https://docs.nextstrain.org/en/latest/reference/snakemake-style-guide.html#log-standard-out-and-error-output-to-log-files-and-the-terminal) with related discussion in https://github.com/nextstrain/docs.nextstrain.org/issues/250. Gist is `exec` makes it easier to capture output for multiple piped commands and we're using it as a standard to easily extend commands in the shell block.

ingest/rules/nextclade.smk
joverlee521: Ah, that's fair. I'll update the intermediate to `subset_nextclade.tsv`.

ingest/defaults/config.yaml
joverlee521: Yeah, I'd be curious to hear what columns from the other segments we should include...would we just want the `qc.overallStatus` for other segments that don't have clade annotations?

ingest/rules/nextclade.smk
huddlej: Classic Tom. Thanks!

ingest/rules/nextclade.smk
joverlee521: Added in 269c99ec9639b1c2b8bfbc8b02abf4423f71bb66

ingest/rules/nextclade.smk
joverlee521: Updated in d0008bfc403582b06fec678fd2d94f7ce696b6fc

ingest/rules/nextclade.smk
joverlee521: Added upload of nextclade.tsv in 1ad1aaec948e7387a8b931a94c5f3ea93b283b18

joverlee521:
Merging per [discussion on Slack](https://bedfordlab.slack.com/archives/C03KWDET9/p1776786137146619). Will address outstanding tasks in separate downstream PRs.

PR #314, opened 2026-04-22
Ingest: More Nextclade runs

joverlee521:
## Description of proposed changes

Now includes all available Nextclade datasets and segments, which is a total of 49 Nextclade runs per ingest. 
Moves the automated ingest workflow to run on AWS Batch to speed up the workflow with more cores. 

## Related issue(s)

Part of https://github.com/nextstrain/seasonal-flu/issues/313

## Checklist

- [x] Checks pass
- [x] Update changelog
- [x] Test [aws-batch workflow](https://github.com/nextstrain/seasonal-flu/actions/runs/25137176310)

Issue #313, opened 2026-04-21
Follow up ingest nextclade


joverlee521:
Issue for tracking follow ups to https://github.com/nextstrain/seasonal-flu/pull/310

- [ ] https://github.com/nextstrain/seasonal-flu/pull/314
    - Move ingest to AWS Batch to speed up workflow with additional cores
    - [Confirm desired Nextclade columns](https://github.com/nextstrain/seasonal-flu/pull/310/commits/a3dc76b499b6f7f3d64d3703b2d247738aecf297#r3061265989)
    - [Confirm desired Nextclade datasets](https://github.com/nextstrain/seasonal-flu/pull/310/commits/a3dc76b499b6f7f3d64d3703b2d247738aecf297#r3061281170)
- [ ] https://github.com/nextstrain/forecasts-flu/pull/42
- [ ] https://github.com/nextstrain/seasonal-flu/pull/315
    - Remove [Nextclade rules](https://github.com/nextstrain/seasonal-flu/blob/master/workflow/snakemake_rules/select_strains.smk#L120-L227) from phylo workflow
    - Remove [emerging haplotype rules](https://github.com/nextstrain/seasonal-flu/blob/d95f3212b0d5f1894bc39eef3ca2725b619c1a49/workflow/snakemake_rules/core.smk#L428-L450) from phylo workflow
    - Remove [derived haplotype rules](https://github.com/nextstrain/seasonal-flu/blob/d95f3212b0d5f1894bc39eef3ca2725b619c1a49/profiles/nextflu-private/report.smk#L113-L146) from phylo workflow
    - update example data to include nextclade + haplotype columns
- [ ] https://github.com/nextstrain/seasonal-flu/issues/308
- [ ] more robust segment checking for `link-gisaid-metadata-and-fasta` ([Slack](https://bedfordlab.slack.com/archives/C03KWDET9/p1776184158401409?thread_ts=1775078611.483489&cid=C03KWDET9))

PR #315, opened 2026-04-25
phylo: Clean up Nextclade and haplotypes

### LINKED ISSUE #308: Move Quickstart GISAID steps to ingest
### Context
This came up in discussion with @huddlej on 1 April 2026.

There will be a lot of duplicate steps between ingest and the phylo workflow as part of #288. The easiest way to simplify the phylo workflow is to move these to the ingest workflow, which already supports directly curating downloaded files from GISAID (which would resolve https://github.com/nextstrain/seasonal-flu/issues/190). 

### Description
TKTK

joverlee521:

_Depends on https://github.com/nextstrain/seasonal-flu/pull/314_

Clean up rules for Nextclade and haplotype annotations since these are now available via ingest. 
Updates the Quickstart for GISAID to reflect these changes. 

## Related issue(s)

Part of https://github.com/nextstrain/seasonal-flu/issues/313
Resolves https://github.com/nextstrain/seasonal-flu/issues/308

## Checklist

- [x] Checks pass
- [x] Update changelog
- [x] Update example data


profiles/example/builds.yaml
```diff
@@ -22,7 +22,6 @@ builds:
       annotation: "config/h3n2/{segment}/genemap.gff"
       tree_exclude_sites: "config/h3n2/{segment}/exclude-sites.txt"
       clades: "config/h3n2/ha/clades.tsv"
-      emerging_haplotypes: "example_data/haplotypes.tsv"
```
joverlee521, line 25: In comparing old and new build with example profile, the old build had the "Subclade" coloring and the new build does not. The "Subclade" coloring in the old build is unexpected since this config does not define the `subclades` config.

joverlee521: OH, the "Subclade" coloring was coming from the merged Nextclade column. So if the build uses the `subclade` config param then "Subclade" used the `augur clades` output, but otherwise it used the Nextclade `subclade` column. I'm going to consider this a "bug" in the old build and not worry about it here.

ingest/README.md
```diff
@@ -13,40 +13,29 @@ own data.
...
-  h1n1pdm: ~
-  vic: ~
-  yam: ~
-  b: ~
-  h1n1: ~
-  h2n2: ~
-  avian-flu: ~
-```
+If your downloaded data only includes a subset of segments and lineages,
+e.g. only include HA segment for h3n2, then you can providing a custom config
```
huddlej:
```suggestion
e.g. only include HA segment for H3N2, then you can providing a custom config
```

profiles/example/builds.yaml
huddlej: Good catch. I never planned for that behavior intentionally, but it's a nice fallback...

ingest/build-configs/gisaid/config.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - ha
+
+filtering:
+  h1n1pdm: ~
+  vic: ~
+  yam: ~
+  b: ~
+  h1n1: ~
+  h2n2: ~
+  avian-flu: ~
```
huddlej, line 29: What if we included comments to describe how these top-level fields work? Or what if we provided the full list of segments and lineages but commented out those we want for the GISAID quickstart? I'm trying to think how an external user would learn what the options are to exclude/include if we ever change the defaults.

ingest/README.md
huddlej: Minor nit, but it made me wonder if you wanted to match the casing for segments and lineages to the ingest config file or to how people would write them out in a human communication?

README.md
```diff
@@ -67,22 +68,47 @@ From the "Download" window that appears, select "Isolates as XLS (virus metadata
...
 Your settings should look like those shown in the screenshot below.
 
 ![Download sequences](images/04-download-sequences.png)
 
 Select the "Download" button.
-Save the FASTA file you downloaded (e.g., `gisaid_epiflu_sequences.fasta`) as `data/h3n2/raw_sequences_ha.fasta`.
+Save the FASTA file you downloaded `ingest/data/<YYYY-MM-DD-N>-sequences.fasta`.
+
+Run the Nextstrain ingest workflow for these data to produce the clean input for
+the phylogenetic workflow with the following command.
```
huddlej, line 98: We might need to tell users to look at this configuration file, so they know they will need to edit it before running ingest with other subtypes or segments. What if we tell them to `cat` the file as part of the quickstart? Then if the file includes the comments and commented lines I mentioned for that config file, users will have a better sense of how the system works.

huddlej (commented):
This looked good to me. I pushed a couple of small changes and left minor comments below. I'm running the full nextflu-private workflow locally now, just to make sure all the reporting parts look good. I'll report back on that tomorrow.

example_data/h3n2_metadata.tsv
```diff
@@ -1,16 +1,16 @@
...
-A/Tehran/1219/2012	flu	EPI446792	2012-01-28	2016-09-03	West Asia	Iran	Tehran	Tehran	undetermined	?	Other Database Import	?	?	True	True	EPI446793	False	False
-A/Victoria/208/2009	flu	EPI272062	2009-06-02	2016-09-03	Oceania	Australia	Victoria	Victoria	cell	Victorian Infectious Diseases Reference Laboratory 	Who Collaborating Centre For Reference And Research On Influenza	?	?	True	False	N/A	False	True
-A/Victoria/210/2009	flu	EPI269899	2009-06-02	2016-09-03	Oceania	Australia	Victoria	Victoria	cell	Victorian Infectious Diseases Reference Laboratory 	Who Collaborating Centre For Reference And Research On Influenza	?	?	True	False	N/A	False	True
-A/Wisconsin/67/2005	flu	EPI367110	2005-XX-XX	2016-09-03	North America	Usa	Wisconsin	Wisconsin	cell	?	Other Database Import	?	?	True	True	EPI367112	False	True
+strain	virus	accession_ha	date	date_submitted	region	country	division	location	passage_category	originating_lab	submitting_lab	age	gender	ha	na	accession_na	gihsn_sample	is_reference	subclade_nextclade_ha	qc.overallStatus_ha	emerging_haplotype_ha	subclade_haplotype_ha	subclade_nextclade_na	qc.overallStatus_na
```
huddlej, line 1: The derived haplotypes here looked unusual (way more amino acid changes relative to "unassigned" clade). I see from [this commit](https://github.com/nextstrain/seasonal-flu/pull/315/changes/841bf34dfc02a7fd4d46a4abea957a0fb78fcdd4) that these are outputs from the new ingest. Since the example data are so much older, their clade annotations will only make sense when the `dataset_name` for `h3n2` and `ha` is set to `flu_h3n2_ha_broad`.

@joverlee521 Would you mind rebuilding this metadata file with that temporary change to the Nextclade config in the new ingest? The new metadata file should look a little simpler.

huddlej:
@joverlee521 The nextflu-private builds looked mostly good. I pushed a couple small fixes to the reporting logic and fixes to emerging haplotypes that I spotted while checking the sanity of the clade labels.

The only outstanding change I'd recommend is the example metadata fix above. Otherwise, this seems good to merge.

profiles/example/builds.yaml
joverlee521: Yeah, now there's no fallback built into the workflow. It depends on the users's metadata including a "subclade" column or the `augur clades` output.

example_data/h3n2_metadata.tsv
joverlee521: Ah, good to know. Thanks for flagging, I'll regenerate the data

ingest/README.md
joverlee521: Ah, the README should probably match human communication. I'll update.

ingest/build-configs/gisaid/config.yaml
joverlee521: I was thinking of adding more docs to the default config, but there's so much going on there that it'll just get lost. I'll add docs here and I like the idea of commented out segments/lineages we want for the quickstart.

ingest/README.md
joverlee521: Updated in https://github.com/nextstrain/seasonal-flu/commit/fd0e01a06bbeba487dcca13bc66294c5895fc0b7

https://github.com/nextstrain/seasonal-flu/blob/fd0e01a06bbeba487dcca13bc66294c5895fc0b7/ingest/README.md?plain=1#L34

ingest/build-configs/gisaid/config.yaml
joverlee521: Added docs to the [gisaid/config.yaml](https://github.com/nextstrain/seasonal-flu/blob/fd0e01a06bbeba487dcca13bc66294c5895fc0b7/ingest/build-configs/gisaid/config.yaml). Maybe a little confusing, users need to uncomment to include segments and comment to include lineages.

README.md
joverlee521: I decided to match the existing [quickstart docs on configs](https://github.com/nextstrain/seasonal-flu/blob/fd0e01a06bbeba487dcca13bc66294c5895fc0b7/README.md?plain=1#L130-L133) and added

https://github.com/nextstrain/seasonal-flu/blob/fd0e01a06bbeba487dcca13bc66294c5895fc0b7/README.md?plain=1#L104-L109

example_data/h3n2_metadata.tsv
joverlee521: I regenerated the data with `flu_h3n2_ha_broad` and pushed up the changes in 2d4ea7642240afe1a169685e4d64fe658abd29bf. This only changed a couple `emerging_haplotype` annotations, but did not actually update the derived haplotypes. 

For reference, this is the config I set for Nextclade
```yaml
nextclade:
  h3n2:
    ha:
      dataset_name: flu_h3n2_ha_broad
      derived_haplotypes:
        genes:
          - HA1
        clade_column: clade
        mutations_column: "founderMuts['clade'].aaSubstitutions"
        derived_haplotype_column: subclade_haplotype_ha
      emerging_haplotypes:
        haplotypes: ../example_data/haplotypes.tsv
        clade_column: clade
        emerging_haplotype_column: emerging_haplotype_ha
      id_field: seqName
      field_map:
        seqName: seqName
        clade: subclade_nextclade_ha
        qc.overallStatus: qc.overallStatus_ha
```

joverlee521: Okay, I can get simpler derived haplotype annotations if I use the first version of the broad dataset. Pushed up in b6cfc55e9a39190788a1640ff75282cb0032b6f5

huddlej: Thanks, @joverlee521! The changes in that last commit look great.

ingest/build-configs/gisaid/config.yaml
huddlej: Awesome! I agree it's odd, but that's a root issue with the config override implementation. This should be enough for most users to get what they need.

huddlej (approved):

PR #323, opened 2026-06-11
Ingest open

### LINKED ISSUE #265: Ingest from INSDC
## Context

[Avian influenza analyses ingest INSDC data](https://github.com/nextstrain/avian-flu/tree/master/ingest/build-configs/ncbi). Additionally, [the Oropouche ingest](https://github.com/nextstrain/oropouche/tree/main/ingest) includes [logic to group multiple segments of the same isolate](https://github.com/nextstrain/oropouche/blob/main/ingest/scripts/group_segments.py).

[GenSpectrum maintains curated INSDC data for seasonal influenza](https://loculus.genspectrum.org/h3n2/search?orderBy=ncbiReleaseDate&page=1&order=descending) which might be a better starting point for our ingest.

## Description

Seasonal influenza should also have an ingest workflow for INSDC data that will:

1. support all segments from H1N1pdm, H3N2, and Vic.
2. provide the standard metadata TSV per isolate and FASTA files per isolate and gene
3. provide Nextclade annotations and alignments

joverlee521:

Add ingest-open workflow to pull h1n1pdm, h3n2, and vic from GenSpectrum and produces 1 metadata.tsv + 8 segment FASTAs per lineage. 

ingest-open is runnable with `nextstrain build` or `nextstrain run`. 

### Test ingest-open with 
1. `nextstrain setup seasonal-flu@ingest-open`
2. `nextstrain run seasonal-flu@ingest-open ingest-open <analysis-dir>`

## Related issue(s)

Resolves https://github.com/nextstrain/seasonal-flu/issues/265

## Checklist

- [x] Checks pass
- [x] Update changelog
- [x] [Test run](https://github.com/nextstrain/seasonal-flu/actions/runs/28206109462)


ingest-open/rules/fetch.smk
```diff
@@ -0,0 +1,142 @@
...
+
+    See <https://api.loculus.genspectrum.org/b-victoria/swagger-ui/index.html#/lapis-controller/getDetailsAsCsv>
+    """
+    endpoint = f"{_genspectrum_lapis_url(wildcards)}/sample/details"
+    params = {
+        "dataFormat": "TSV",
+        "versionStatus": "LATEST_VERSION",
+        "compression": "zstd",
+        "fields": config["genspectrum_metadata_fields"],
+    }
```
joverlee521, line 48: OH, just realized this includes non-human host data. We can probably filter by host field here.

huddlej: I was just going to comment that `h1n1pdm` records include 9305 samples from pigs (e.g., A/swine/Illinois/A03690537/2026). In the long term, we should support multiple hosts (at least including a `host` column in the metadata for builds to filter on), but for now we could just filter to human hosts.

ingest-open/scripts/parse-genspectrum-division
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+"""
+Parse Pathoplexus division field into two separate fields, "division" and "location".
```
huddlej:
```suggestion
Parse GenSpectrum division field into two separate fields, "division" and "location".
```

ingest-open/defaults/config.yaml
```diff
@@ -0,0 +1,299 @@
...
+
+segments:
+  - pb2
+  - pb1
+  - pa
+  - ha
+  - np
+  - na
+  - mp
+  - ns
```
huddlej, line 16: If we want to define a custom ingest config (H1N1pdm only, HA/NA only), do we need to use `nextstrain build` with a local copy of the workflow? I don’t see a way to pass a custom config file path to Snakemake through `nextstrain run` (probably by design!).

ingest-open/rules/fetch.smk
joverlee521: Added filter to human host and the `host` field in metadata in 4f968a0

profiles/example-open/builds.yaml
```diff
@@ -0,0 +1,33 @@
+lat-longs: "config/lat_longs.tsv"
+
+filter:
+  metadata-id-columns: accession
```
huddlej: I had to add this option and corresponding Snakemake logic to fix the issue where sequences use accessions as ids and metadata used strain names as ids by default.

I haven't updated every place in the workflow where `--metadata-id-columns` would need to be added, since this may not be the best solution.

ingest-open/rules/curate.smk
```diff
@@ -0,0 +1,227 @@
...
+        ndjson="data/{lineage}/open.ndjson.zst",
+        geolocation_rules=resolve_config_path(config["curate"]["local_geolocation_rules"]),
+        annotations=resolve_config_path(config["curate"]["annotations"]),
+    output:
+        ndjson="data/{lineage}/curated.ndjson.zst",
+    params:
+        field_map=format_field_map(config["curate"]["field_map"]),
+        strain_regex=config["curate"]["strain_regex"],
+        strain_backup_fields=config["curate"]["strain_backup_fields"],
+        date_fields=config["curate"]["date_fields"],
```
huddlej, line 36: A little less than 10% of each subtype’s records have ambiguous dates (5520 of 63473 H3N2, 4171 of 51238 H1N1pdm, 2118 and 21295 Vic). Some of these records are from submitters we work with (e.g., [Influenza A virus \(A/Z6RHY1/USA/2022\(H1N1\)\(H1N1\)\) segment 1 polymerase - Nucleotide - NCBI](https://www.ncbi.nlm.nih.gov/nuccore/PX732762.1/)) and who could potentially update records to have more detailed dates. This isn't something we can address in this PR, but I wanted to document the issue somewhere.

huddlej (commented):
The ingest workflow ran without any issues and the phylo workflow mostly worked as expected. I pushed a couple of fixes to the phylo workflow, but the main blocking issue is the one of `accession` vs. `strain` as the id for sequences and metadata. I pushed a WIP commit that allows users to set the `--metadata-id-columns` from their build config instead of renaming the `accession` column to `strain` in the metadata. I didn't update all places in the workflow where we'll need that custom argument, but I can if that seems like the right solution.

A non-blocking issue for open ingest but something to consider for future work: I noticed a lot of output gets written to stdout and the various logs, but unless I know what I’m looking for I’ll probably never dig into those logs. For people who run ingest, we could make a summary report at the end of the workflow that includes the total number of records per lineage/segment from NCBI, the number that successfully aligned with Nextclade and have “good” QC, etc. It looks like we could make this report from a single rule and all the intermediate/final files the workflow produces. This kind of report would help us catch silent failures in the workflow when the numbers don't match our expectations.

ingest-open/defaults/config.yaml
joverlee521: The workflow automatically uses a config if it's available in the analysis directory 

https://github.com/nextstrain/seasonal-flu/blob/88875837eff5dd1724951dbaa737ff97583eb686/ingest-open/Snakefile#L13-L15

So all you need to do is add the custom config to `<analysis_dir>/config.yaml`. I can add this to the README usage docs.

profiles/example-open/builds.yaml
joverlee521: Ah, this is a bug in the ingest-open workflow! The sequences should be using `strain` since we dedup by strain. I'll push up a fix.

ingest-open/defaults/config.yaml
huddlej: Ooh cool! In this use case, I would only want to override a subset of the default config without needing to copy all of the curate settings, etc. But this starts to get into the rabbit hole of how we merge config files in workflows...I don't want to block this on that, so copying the whole config and modifying locally is good.

profiles/example-open/builds.yaml
huddlej: Thank you! You can drop my WIP commits when you push the fix, since those shouldn't be necessary anymore

joverlee521: Aaaand of course after trying to change it to `strain`, I remember why I used `accession`. 

The `strain` field from open data if pretty messy including characters which are not accepted by Augur. Instead of trying to clean up the `strain` values, other pathogens like mpox uses the [accession as the id field](https://github.com/nextstrain/mpox/blob/edbd493e4ab05c0187bf2c7d288cfec637c83d93/phylogenetic/defaults/mpxv/config.yaml#L19) and then uses [`strain` as the default tip label](https://github.com/nextstrain/mpox/blob/edbd493e4ab05c0187bf2c7d288cfec637c83d93/phylogenetic/defaults/mpxv/auspice_config.json#L108). 

So I'm think I'm going to build on your WIP commit and make sure that `--metadata-id-column` is used where appropriate.

joverlee521: WIP 990096d for adding `--metadata-id-column`. I realized in the process how much the custom scripts in this repo use `strain` so those will all need to be updated, which I'll dig more into next week.

ingest-open/defaults/config.yaml
joverlee521: Ah, the custom config does get merged on top of the default config by Snakemake's config merging, so you should be able to override a subset of params

profiles/example-open/builds.yaml
joverlee521: After our chat this week, I've moved the open build changes to a downstream branch ([phylo-open](https://github.com/nextstrain/seasonal-flu/tree/phylo-open)) so that I can merge this PR and we can tackle the phylo changes separately.

joverlee521: Tracking tasks in https://github.com/nextstrain/seasonal-flu/issues/330.

joverlee521:
I'll merge this on Monday and we can continue to make updates to ingest-open later based on whatever changes we need for #330.

huddlej (approved):

PR #327, opened 2026-06-17
Ingest: exclude records

### LINKED ISSUE #326: ingest: Add ability to exclude records
### Context

I worked around the csv error in #324, but looks like a particular record has been causing issues for downstream users as reported on Slack [[1](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1781641095340389), [2](https://bedfordlab.slack.com/archives/C03KWDET9/p1781707169749309?thread_ts=1781706824.756529&cid=C03KWDET9)]. We can't exclude the record in the phylo workflow because it leads to the same csv error in `augur filter`. We could [increase the field limit in Augur](https://github.com/nextstrain/augur/issues/2013), but we should also be able to exclude bad records in ingest. 

### Possible solutions

1. Filter out the record in [concatenate_gisaid_ndjsons](https://github.com/nextstrain/seasonal-flu/blob/af1fb3904767333cccfff2bf49f1cd26cddd2b3b/ingest/rules/prepare_ndjson.smk#L71-L85). The record will be filtered out once and never included in the NDJSON cache or the downstream outputs. 
2. Filter out the record at the start of [curate](https://github.com/nextstrain/seasonal-flu/blob/af1fb3904767333cccfff2bf49f1cd26cddd2b3b/ingest/rules/curate.smk#L42). The record will still be in the NDJSON cache but filtered out from downstream outputs. 

I'm going with [2] because this will allow us to "undo" the exclusion easily.

joverlee521:

Add step to filter out records by id so that we can exclude "bad" records from the curated output. Excludes a single record
based on discussion in [Slack](https://bedfordlab.slack.com/archives/C03KWDET9/p1781706824756529).

## Related issue(s)

Resolves https://github.com/nextstrain/seasonal-flu/issues/326

## Checklist

- [ ] Checks pass
- [ ] Update changelog

### DIFF
--- a/ingest/curation-docs.md
+++ b/ingest/curation-docs.md
@@ -16,6 +16,10 @@ I think the experience goes something like:
 3. Make code/TSV changes, following instructions here
 4. Rerun the curate pipeline & diff against (2). We want a user-friendly diffing approach which identifies what's changed without being too verbose
 
+## Exclude records
+
+If you run into "bad" records that you want to exclude entirely from the outputs,
+then add their GISAID EPI ISL to the `defaults/exclude.txt` file.
 
 ## Geography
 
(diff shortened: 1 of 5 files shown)


joverlee521:
I'll merge this tomorrow before I run the weekly ingest.

Issue #330, opened 2026-06-26
Add open builds


joverlee521:
With https://github.com/nextstrain/seasonal-flu/pull/323, we will have public data to use for open builds. I started on an example open build when I was working on the ingest and is now on the [phylo-open branch](https://github.com/nextstrain/seasonal-flu/tree/phylo-open).

### TODOs
- [x] https://github.com/nextstrain/seasonal-flu/pull/331
- [x] https://github.com/nextstrain/seasonal-flu/pull/338
- [ ] https://github.com/nextstrain/seasonal-flu/pull/332
- [ ] https://github.com/nextstrain/seasonal-flu/pull/346
- [ ] https://github.com/nextstrain/seasonal-flu/pull/348
- [ ] https://github.com/nextstrain/nextstrain.org/pull/1374

PR #331, opened 2026-06-26
Support multiple inputs

joverlee521:

Optionally use the `inputs` param to define multiple inputs for the workflow, backwards compatible with existing profiles. Mainly motivated by the need for defining inputs for the forthcoming open builds, but it's nice to support multiple inputs!

Note this uses the latest `augur merge` features that requires Augur v34.0.0 as part of https://github.com/nextstrain/public/issues/44. It deviates from other pathogens to use "strain" as the merged metadata id field since so many of downstream scripts hardcode "strain". 

## Related issue(s)

Part of https://github.com/nextstrain/seasonal-flu/issues/330
Part of https://github.com/nextstrain/public/issues/44

## Checklist

- [x] Checks pass
- [x] Update changelog


workflow/snakemake_rules/merge_inputs.smk
```diff
@@ -139,8 +153,10 @@ rule merge_metadata:
 
         augur merge \
             --metadata {params.metadata:q} \
+            --metadata-id-columns {params.id_field:q} \
             --source-columns 'input_{{NAME}}' \
-            --output-metadata {output.metadata}
+            --output-metadata {output.metadata} \
+            --output-metadata-id-column strain
```
joverlee521, line 159: Hah, I thought this would be a way for me to use `accession` as the `id_field` for the open data, but this runs into an error because the metadata also has `strain` column:

```
ERROR: Non-id column names in metadata inputs may not conflict with the
output id column name ('strain', the given --output-metadata-id-column).

The following input column would conflict:

  'strain' in metadata table 'example' (id column: 'accession')

Please rename or drop the conflicting column before merging.
Renaming may be done with `augur curate rename`.
```
Briefly thought about updating `augur merge` to change the conflicting column name, but this was [shot down in review of the original PR](https://github.com/nextstrain/augur/pull/1593#discussion_r1725737170)

victorlin: If `augur merge` renamed the existing `strain` to `__strain`, would it still be useful later?

joverlee521: Good question! I'm trying to figure out what behavior would be appropriate here...

In this particular case, I'd want the field as a metadata column to export so it's displayed in the tree tips. I'd probably want to rename it to `display_strain` or `original_strain`. This isn't necessarily true for all data sources, so it's more appropriate for me to do upstream of the merge.

huddlej: I never feel good about modifying the original metadata to make it work for our workflows, since it would infuriate me as an external user to have that happen to my own data ("I _need_ that `strain` column for something else!"). But if we're going to do it for these open data, I'd prefer `original_strain` or maybe `genbank_strain`. The `display_strain` option suggests that we want to use the contents for display somewhere which we may really not if the contents are unreliable.

workflow/snakemake_rules/merge_inputs.smk
```diff
@@ -0,0 +1,183 @@
+"""
+This part of the workflow merges inputs based on what is defined in the config.
+
+OUTPUTS:
+
+    metadata  = "data/{lineage}/metadata.tsv"
+    sequences = "data/{lineage}/{segment}.fasta"
+
```
huddlej, line 8: This documentation is great, @joverlee521! How would you feel about moving/copying it into the README? This workflow is already under-documented such that users would have no idea what functionality exists without digging into our own build configs or the code. Maybe these docs could be the seed of proper documentation for the whole workflow?

Ideally, we'd have more formal docs like the ncov repo, but not in this PR! 😅

joverlee521: Yes! I can add more docs on this in the README. Reading over it now reminded me that I can probably update the GISAID Quickstart to use this functionality as well!

workflow/snakemake_rules/merge_inputs.smk
joverlee521: Same, I'm also not a fan of modifying the original strain name but I think unavoidable for deduplication purposes. 

Downstream of this PR, I'm going to modify ingest-open to clean up the strain names into the `strain` column and use it as the id since the GenSpectrum accessions are not reliable. I'll keep the original value as a separate `original_strain` column for display purposes.

workflow/snakemake_rules/merge_inputs.smk
joverlee521: Add config section to README in [1dd1813](https://github.com/nextstrain/seasonal-flu/pull/331/commits/1dd18136d4e457c20e89e75ccef81638a46e0329)

workflow/snakemake_rules/merge_inputs.smk
joverlee521: Modifying ingest-open in https://github.com/nextstrain/seasonal-flu/pull/338.

PR #332, opened 2026-06-26
Add nextstrain-open build

joverlee521:

Add example-open and nextstrain-open configs to use the open data. 

## Related issue(s)

Part of https://github.com/nextstrain/seasonal-flu/issues/330

## Checklist

- [x] Checks pass
- [x] [Test run](https://github.com/nextstrain/seasonal-flu/actions/runs/31822420912) -> https://nextstrain.org/staging/seasonal-flu/open/h3n2/ha/2y
- [ ] Update TBD in changelog


profiles/nextstrain-open.yaml
```diff
@@ -0,0 +1,136 @@
...
+      reference: "config/{lineage}/{{segment}}/reference.fasta"
+      annotation: "config/{lineage}/{{segment}}/genemap.gff"
+      tree_exclude_sites: "config/{lineage}/{{segment}}/exclude-sites.txt"
+      clades: "config/{lineage}/ha/clades.tsv"
+      subclades: "config/{lineage}/{{segment}}/subclades.tsv"
+      auspice_config: "config/{lineage}/{{segment}}/auspice_config.json"
+      vaccines: "config/{lineage}/vaccine.json"
+      enable_glycosylation: true
+      enable_lbi: true
+      enable_embeddings: true
```
joverlee521: [Test run in GH Action](https://github.com/nextstrain/seasonal-flu/actions/runs/31546633554) errored on `embed_with_tsne`

```
Traceback (most recent call last):
  File "/usr/local/bin/pathogen-embed", line 8, in <module>
    sys.exit(run_embed())
             ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pathogen_embed/__main__.py", line 109, in run_embed
    return embed(args)
           ^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pathogen_embed/pathogen_embed.py", line 565, in embed
    embedding = embedder.fit_transform(distance_matrix)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sklearn/utils/_set_output.py", line 295, in wrapped
    data_to_wrap = f(self, X, *args, **kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sklearn/base.py", line 1474, in wrapper
    return fit_method(estimator, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/sklearn/manifold/_t_sne.py", line 1135, in fit_transform
    self._check_params_vs_input(X)
  File "/usr/local/lib/python3.11/site-packages/sklearn/manifold/_t_sne.py", line 846, in _check_params_vs_input
    raise ValueError("perplexity must be less than n_samples")
ValueError: perplexity must be less than n_samples
```

huddlej: pathogen-embed should handle that error more gracefully, but the default perplexity is 30. Do you expect to have fewer than 30 samples in the open build? Maybe for a 6-month build there isn't that much more data?

[The workaround is to set the perplexity to a smaller value](https://github.com/nextstrain/seasonal-flu/blob/4962eeb1ef80f76ccf1ab3eac8ba2a6300eecded/profiles/ci/builds.yaml#L23-L25). This is a global setting right now, so it would change embeddings for all the builds. Probably the better short-term fix is to disable embeddings like you had before.

joverlee521: > Do you expect to have fewer than 30 samples in the open build? Maybe for a 6-month build there isn't that much more data?

Hmm, there should be more than 30 samples in the 6m build, but yeah I'll turn off embeddings for now.

profiles/nextstrain-open/overwrite_auspice_config.smk
```diff
@@ -0,0 +1,16 @@
...
+  in Augur does not support removing elements from the base list.
+  <https://github.com/nextstrain/augur/blob/c8cc27cade9268fbbeec806f0e7ba5b7ba45c649/augur/util_support/auspice_config.py#L44-L57>
+  """
+  input:
+    auspice_config = "config/{lineage}/{segment}/auspice_config.json",
+  output:
+    auspice_config = "data/{lineage}/{segment}/auspice_config.json",
+  shell:
+    """
+    jq '.data_provenance = [{{"name": "GenSpectrum", "url": "https://loculus.genspectrum.org"}}]' \
```
huddlej: Instead of special-casing the provenance of these builds, what if we either:

1. add a rule to the main workflow that always adds a provenance based on values defined in the build config (e.g., `data_provenance: 'gisaid'`), so the provenance becomes an explicit configuration parameter instead of implicitly defaulting to one provenance

or

2. add an argument to `augur export v2` that allows the user/workflow to specify the data provenance as an argument that overrides the config JSON instead of unioning with the config (`--data-provenance '[{{"name": "GenSpectrum", "url": "https://loculus.genspectrum.org"}}]'`) and make the export rule in this workflow pull the provenance from the build config

or

3. make copies of the configs for the open data like `config/{lineage}/{segment}/auspice_config_{provenance}.json`

I get that this last option is what you're trying to avoid with this custom rule.

profiles/nextstrain-open/description.md
```diff
@@ -0,0 +1,8 @@
+We gratefully acknowledge the authors, originating and submitting laboratories
```
huddlej, line 1: What if the description included links to the URLs that we provide for all open flu data like? The list wouldn't have to be as complex as [the ncov open description's](https://github.com/nextstrain/ncov/blob/5949c4044edb3c25ab9bce03286bac8d5b1ce337/nextstrain_profiles/nextstrain-open/nextstrain_description.md). At the least it could be a link to https://nextstrain.org/pathogens/files?filter=seasonal-flu

README.md
```diff
@@ -135,6 +135,16 @@ To skip subsampling and use all records that you downloaded from GISAID, set `fi
             filters: ""
 ```
 
+## Open build
+
+The open build uses public data from GenSpectrum, which we curate and make publically available.
```
huddlej: I'd say these public data are from INSDC databases and initially curated by GenSpectrum.

profiles/nextstrain-open.yaml
```diff
@@ -0,0 +1,139 @@
...
+    build_params: &shared-build-params
+      auspice_name: "seasonal-flu_open_{lineage}_{{segment}}_{resolution}"
+      reference: "config/{lineage}/{{segment}}/reference.fasta"
+      annotation: "config/{lineage}/{{segment}}/genemap.gff"
+      tree_exclude_sites: "config/{lineage}/{{segment}}/exclude-sites.txt"
+      clades: "config/{lineage}/ha/clades.tsv"
+      subclades: "config/{lineage}/{{segment}}/subclades.tsv"
+      # Created in the workflow with custom rule `overwrite_auspice_config`
+      auspice_config: "data/{lineage}/{{segment}}/auspice_config.json"
+      vaccines: "config/{lineage}/vaccine.json"
```
huddlej, line 65: Note to self to check that the vaccine strains actually appear in the open data with the same names.

profiles/nextstrain-open.yaml
```diff
@@ -0,0 +1,139 @@
...
+      enable_glycosylation: true
+      enable_lbi: true
+      # Turn off embeddings for now since it errors for 6m builds
+      # See <https://github.com/nextstrain/seasonal-flu/pull/332#discussion_r3768509224>
+      enable_embeddings: false
+      include: "'config/{lineage}/reference_strains.txt'"
+      exclude: "'config/{lineage}/outliers.txt'"
```
huddlej, line 72: We may need to define new reference and outlier files for open data or make new entries in the existing files.

profiles/nextstrain-open.yaml
huddlej: I found that several H3N2 vaccine strains didn't have exact matches among the strain names in the open metadata (e.g., none of the Swiss strains post-2012, the recent Croatia strain, etc.). We will need to look for corresponding records in GenBank for all of these missing strains across all three subtypes.

profiles/nextstrain-open.yaml
huddlej: I ended up adding new entries to the existing files during my testing.

huddlej (approved):
This worked well for me out of the box, @joverlee521. I made my own build covering all time for H3N2 and H1N1pdm which forced me to play around with the settings more. Everything worked as expected.

I had a question/suggestions about how we handle Auspice config JSON overrides, but I know that's a sticky subject better had with the rest of the team.

The main blocker I could imagine to releasing open builds based on this is that not all of the vaccine strains from our original builds have matching records in the open data. This is something we could look into outside of this PR, though.

profiles/nextstrain-open/overwrite_auspice_config.smk
joverlee521: Ah thanks for flagging! This was a quick workaround for the `example-open` build, so I should definitely revisit. 
Yes, definitely trying to avoid (3), so I think (1) is the most appealing. 

This also reminds me that we should probably rename the GISAID builds if we are going to move to `seasonal-flu/gisaid/...` and `seasonal-flu/open/...` URLs.

profiles/nextstrain-open/description.md
joverlee521: Added in 37c0741152f118858a176476eb62c889d1e45f0e

README.md
joverlee521: Reworded in 37c0741152f118858a176476eb62c889d1e45f0e

profiles/nextstrain-open/overwrite_auspice_config.smk
joverlee521: After today's in person discussion, will pivot to layering auspice configs for the separate data provenances. I'll update the PR to define maintainers and data_provenance in separate JSONs that can be layered on top the base segment auspice config.

joverlee521: > I'll update the PR to define maintainers and data_provenance in separate JSONs that can be layered on top the base segment auspice config.

This is now done with https://github.com/nextstrain/seasonal-flu/compare/aa49e1bddabe4087488de25f5272a4a6259ba482...f891b200123c6195888469850f40e7f5e3a00456

PR #345, opened 2026-08-17
Annotate fitness effects for recurrent substitutions based on Turner …

### LINKED ISSUE #344: Annotate fitness effects for recurrent substitutions based on Turner et al. method

SamT123:

Add `lcr_sum`, the sum of LCR values / fitness effects from Turner et al. 2026 [1] for each substitution on the path from the root of the tree to the node, for H3N2 HA tree.

The LCR value for a substitution on a branch is taken from the 1y time interval containing that branch, from https://github.com/SamT123/nextstrain-flu-convergence.

scripts/calculate_lcr_sum.py is based on scripts/calculate_antigenic_advance.py

Substitutions on branches older than the first window are given a value of 0; the value for substitutions on branches after the last window is taken from the last window.

[1] https://www.biorxiv.org/content/10.64898/2026.04.10.717627v1

## Related issue(s)

- Resolves #344

## Checklist

I've run `nextstrain build . --configfile profiles/ci/builds.yaml` locally, and the output appears correct.

- [ ] Checks pass
- [ ] Update changeling

### DIFF
--- a/scripts/calculate_lcr_sum.py
+++ b/scripts/calculate_lcr_sum.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+import argparse
+
+from augur.io import write_json
+from augur.utils import read_node_data, read_tree
+from bisect import bisect_right
+import pandas as pd
+
+
+def annotate_tree(mutations_by_node, tree, weights, starts, ends, dates):
+    """Annotates log convergence ratio attributes to nodes of a given tree.
+    """
+    lcr_sum = {tree.root.name: 0.0}
+    total = 0
+    missing = 0
+    outside_windows = 0
+    after_last_window = 0
+
+    for node in tree.find_clades():
+        for child in node.clades:
+            # Get mutations between current node and its parent.
+            mutations = [
+                (gene, mutation)
+                for gene, gene_mutations in mutations_by_node[child.name].items()
+                for mutation in gene_mutations
+            ]
+
+            # Find the window to use for scoring the current node (earliest end date after the branch date)
+            date = dates[child.name]
+            window = bisect_right(ends, date)
+
+            # Calculate the log convergence ratio sum on the branch to the current node.
+            total += len(mutations)
+            branch = 0.0
+
+            if window == len(ends):
+                # Past every window's end, so score with the last window.
+                after_last_window += len(mutations)
+                window = len(ends) - 1
+
+            if starts[window] > date:
+                outside_windows += len(mutations)
+            else:
+                for gene, mutation in mutations:
+                    weight = weights.get((f"{gene}:{mutation}", window))
+                    if weight is None:
+                        missing += 1
+                        continue
+
+                    branch += weight
+
+            # Calculate cumulative log convergence ratio from root to current node.
+            lcr_sum[child.name] = lcr_sum[node.name] + branch
+
+    return lcr_sum, total, missing, outside_windows, after_last_window
+
+
+def main(args):
+    """Calculate log convergence ratio.
+    """
+    # Load log convergence ratios &  time windows
+    weight_table = pd.read_csv(args.weights, sep="\t", comment="#")
+
+    # Sort windows by end date to help find first window with end date after current node in annotate_tree
+    windows = pd.read_csv(args.windows, sep="\t").sort_values("window_end_numeric", ignore_index=True)
+    starts = windows["window_start_numeric"].tolist()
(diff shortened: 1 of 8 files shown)


scripts/calculate_lcr_sum.py
```diff
@@ -0,0 +1,133 @@
...
+                for gene, mutation in mutations:
+                    weight = weights.get((f"{gene}:{mutation}", window))
+                    if weight is None:
+                        missing += 1
+                        continue
+
+                    branch += weight
+
+            # Calculate cumulative log convergence ratio from root to current node.
+            lcr_sum[child.name] = lcr_sum[node.name] + branch
```
huddlej:
```suggestion
    for node in tree.find_clades():
        # Get mutations between current node and its parent.
        mutations = [
            (gene, mutation)
            for gene, gene_mutations in mutations_by_node[node.name].items()
            for mutation in gene_mutations
        ]

        # Find the window to use for scoring the current node (earliest end date after the branch date)
        date = dates[node.name]
        window = bisect_right(ends, date)

        # Calculate the log convergence ratio sum on the branch to the current node.
        total += len(mutations)
        branch = 0.0

        if window == len(ends):
            # Past every window's end, so score with the last window.
            after_last_window += len(mutations)
            window = len(ends) - 1

        if starts[window] > date:
            outside_windows += len(mutations)
        else:
            for gene, mutation in mutations:
                weight = weights.get((f"{gene}:{mutation}", window))
                if weight is None:
                    missing += 1
                    continue

                branch += weight

        # Calculate cumulative log convergence ratio from root to current node.
        lcr_sum[node.name] = branch
        if getattr(node, "parent"):
            lcr_sum[node.name] += lcr_sum.get(node.parent.name, 0.0)

```

scripts/calculate_lcr_sum.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python3
+import argparse
+
+from augur.io import write_json
+from augur.utils import read_node_data, read_tree
```
huddlej:
```suggestion
from augur.utils import annotate_parents_for_tree, read_node_data, read_tree
```

scripts/calculate_lcr_sum.py
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    # Load node dates.
+    dates = {
+        name: node["numdate"]
+        for name, node in read_node_data(args.branch_lengths)["nodes"].items()
+    }
+
+    # Load tree.
+    tree = read_tree(args.tree)
```
huddlej, line 84:
```suggestion
    tree = read_tree(args.tree)
    tree = annotate_parents_for_tree(tree)
```

huddlej (commented):
Thanks, @SamT123! This looks really good and ran fine with test data and a full Nextstrain 3-year tree. I only had one question about the LCR script.

I noticed that you annotate the tree starting with the children of the root node. Did you intend to exclude the root node from the LCR calculations? A subtle feature of Nextstrain trees is that during `augur export` we annotate a hidden parent node to the MRCA of the tree which represents the alignment reference. We assign the branch between that new reference node and the MRCA with the mutations that exist between the reference and MRCA sequences. Nextclade can use any tree with this reference branch as an ad hoc "Nextclade dataset" which is handy. This also means that the `muts.json` has mutations for the MRCA (usually the node `NODE_0000000`) and that you could calculate the LCR sum starting with the root node of the tree. I proposed some changes to the LCR script below that allow you to include the root node, if you like. Below are before and after scatterplots of LCR by date for a smaller subsampled H3N2 HA tree. You can see that the relative values of the tips all remain the same, but the intercept/root value changes.

Before including root node:

Image

After including root node:

Image

If you prefer the original approach without the root node, you can ignore the code suggestions below.

scripts/calculate_lcr_sum.py
huddlej: This function annotates a `parent` node to each node in the tree. The value of `tree.root.parent` is `None`. When you have these parent pointers, you can traverse the tree in preorder and reference the parent of each node without needing to loop through the children of each node.

SamT123: Ah that's handy!

SamT123:
I'm happy either way re. scoring the substitutions between the ref & the root. It sounds scoring those mutations  matches the pattern you expect & maybe have used elsewhere, so let's do that!

Issue #343, opened 2026-08-05
Add MLR fitness coloring


joverlee521:
From discussion with @huddlej 

- [ ] add coloring for MLR fitness (autogenerated haplotypes? TBD) 
- [ ] orchestrate GH workflows to work in the correct order (ingest -> titers -> forecasts-flu/run-models -> seasonal-flu public + private builds)
    - https://github.com/nextstrain/forecasts-flu/pull/46
    - https://github.com/nextstrain/seasonal-flu/pull/349

PR #349, opened 2026-09-09
.github/upload: Move seasonal-flu build triggers to forecasts-flu

joverlee521:

In order to use latest MLR results as a colorings in the builds, we need to run the builds after the model workflow in forecasts-flu has completed. So instead of triggering the builds directly after the titer step, this now only triggers forecasts-flu/run-models with the inputs to trigger the seasonal-flu builds after the model workflow.

I did not add the option to trigger builds independently of the forecasts-flu workflow because that can be done by manually triggering the respective seasonal-flu build GH Actions directly.

## Related issue(s)

Part of https://github.com/nextstrain/seasonal-flu/issues/343
Depends on <https://github.com/nextstrain/forecasts-flu/pull/46>

## Checklist

- [ ] Checks pass


jameshadfield (commented):
General comment - all of these moving parts are so hard for someone on the periphery of the orchestration layer here. Have you considered something like a mermaid chart, [which are natively rendered in GitHub](https://github.blog/developer-skills/github/include-diagrams-markdown-files-mermaid/)? Here's a very poor 10min attempt, but I'm imagining we could cover everything in the three (or more?) repos, all the way from ingest to uploads



```mermaid
swimlane-beta TB
  subgraph seasonal-flu
    runModels[run-models]
    foo[???]
    bar[???]
  end

  subgraph avian-flu
    something{TKTK}
  end

  subgraph forecast-flu
    run_models[run_models]
    triggerForecastsFlu[? workflow dispatch:\ntriggerForecastsFlu]
    triggerPublic[? workflow dispatch:\ntriggerPublic]
  end

  runModels --> run_models --> triggerForecastsFlu
  run_models --> triggerPublic
  triggerForecastsFlu --> foo
  triggerPublic --> bar

```

README.md
```diff
@@ -191,6 +191,46 @@ inputs:
         na: ingest/new-results/h3n2/na.fasta
 ```
 
+## GitHub Action Workflows
```
joverlee521: > Have you considered something like a mermaid chart, [which are natively rendered in GitHub](https://github.blog/developer-skills/github/include-diagrams-markdown-files-mermaid/)? Here's a very poor 10min attempt, but I'm imagining we could cover everything in the three (or more?) repos, all the way from ingest to uploads

Thanks for the suggestion @jameshadfield! I've added the current interactions of workflows in 106ff8cf92ed71d0ef6f621bcf8954387a33d52c and the updated interactions in 4b518a03ebcd2d867c93f145e449881463a58622.

